PR #7, opened 2023-01-18
Update Agreement Concept with additional attributes

martinhalford:
Signed-off-by: Martin Halford <martin@benext.io>

Add properties/attributes to Agreement Concept in order to better support instantiation of agreements. 

This includes:
- Parties
- State
- History
- Signatures
- Attachments/References
- Metadata

This PR is intended to be a work-in-progress / draft proposal / starting-point upon which to iterate.

### DIFF
--- a/model/protocol.cto
+++ b/model/protocol.cto
@@ -3,7 +3,7 @@ namespace org.accordproject.protocol@1.0.0
 
 import concerto.metamodel@0.4.0.{Property,ConceptDeclaration,Model} from https://models.accordproject.org/concerto/metamodel@0.4.0.cto
 import org.accordproject.commonmark@0.5.0.Document from https://models.accordproject.org/markdown/commonmark@0.5.0.cto
-import org.accordproject.party.* from https://models.accordproject.org/accordproject/party.cto
+import org.accordproject.party@0.2.0.* from https://models.accordproject.org/accordproject/party@0.2.0.cto
 
 scalar JSON extends String
 scalar FullyQualifiedTypeName extends String
@@ -91,9 +91,9 @@ concept Agreement identified by agreementId {
     o Signature[] signatures                // Signatures of the parties to the agreement
     o AgreementStatusType agreementStatus   // Current status of agreement
     o History[] history                     // History of document state and details.
-    o Blob[] attachments optional           // Pdfs, images, multimedia, etc. that form part of the agreement
-    o String[] references optional          // uri/urls to external references relevant to agreement
-    o String[] metadata optional            // Additional data that may be relevant to the agreement
+    o Blob[] attachments                    // Pdfs, images, multimedia, etc. that form part of the agreement
+    o String[] references                   // uri/urls to external references relevant to agreement
+    o String[] metaData                     // Additional data that may be relevant to the agreement
 }
 
 @resource
@@ -112,7 +112,7 @@ participant AgreementParty extends Party {
 @resource
 @description("An Address of an Agreement Party")
 concept Address {
-  o String[] streetRoad optional
+  o String[] streetRoad
   o String suburbTownCity optional
   o String stateTerritoryRegion optional
   o String postalCode optional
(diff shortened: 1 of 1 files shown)


model/protocol.cto
```diff
@@ -3,14 +3,15 @@ namespace org.accordproject.protocol@1.0.0
 
 import concerto.metamodel@0.4.0.{Property,ConceptDeclaration,Model} from https://models.accordproject.org/concerto/metamodel@0.4.0.cto
 import org.accordproject.commonmark@0.5.0.Document from https://models.accordproject.org/markdown/commonmark@0.5.0.cto
+import org.accordproject.party.* from https://models.accordproject.org/accordproject/party.cto
```
dselman, line 6: Should use an explicit (no '*') versioned import.

model/protocol.cto
```diff
@@ -86,7 +87,54 @@ concept Agreement identified by agreementId {
     o String agreementId
     o JSON data
     --> Template template
-    // state?
+    o AgreementParty[] agreementParties     // Parties to the agreement
+    o Signature[] signatures                // Signatures of the parties to the agreement
+    o AgreementStatusType agreementStatus   // Current status of agreement
+    o History[] history                     // History of document state and details.
```
dselman, line 93: optional?

model/protocol.cto
```diff
@@ -86,7 +87,54 @@ concept Agreement identified by agreementId {
...
+    o JSON data                             // Data at time of change
+    o String[] metadata optional            // Additional data that may be relevant to the agreement at time of change
+}
+
+@resource
+@description("A signature of an Agreement Party")
+concept Signature {
+    o AgreementParty signatory              // The Agreement Party signing the Agreement
+    o DateTime signedAt optional            // When the signing occurred
+    o String[] metaData optional            // Geolocation data, IP address, etc.
```
dselman, line 136: differs in case "metaData" from rest.

model/protocol.cto
```diff
@@ -86,7 +87,54 @@ concept Agreement identified by agreementId {
...
+  o String country optional
+}
+
+@resource
+@description("The History of an Agreement")
+concept History {
+    o DateTime dateTime
+    o AgreementStatusType agreementStatus   // Status at time of change
+    o JSON data                             // Data at time of change
+    o String[] metadata optional            // Additional data that may be relevant to the agreement at time of change
```
dselman, line 128: 3x instances of metadata - perhaps this should be a concept?

dselman (changes requested):
Nice!

model/protocol.cto
martinhalford: OK. Setting to `import org.accordproject.party@0.2.0.*`

model/protocol.cto
martinhalford: As `agreementParties`, `signatures`, and `history` are all arrays, and can be empty, then I thought it would be safer & simpler to make them mandatory.  All agreement instances will, ultimately, have `agreementParties`, `signatures`, and some `history`- therefore, not wasted space.

As for `agreementStatus`, I was thinking that any agreement instance must have a status - either `Draft`, `Signing`, `Completed` or `Superseded` - so, personally, think this should be mandatory.  If we come across a use case that is not handled by these 4 status types then we can add it (e.g `None` could be an enum).

martinhalford: Based on this logic, I've removed `optional` from other arrays in the bits I added.  Must admit, `optional` fields have caused me pain in the past - so tend to use them only when I have to. :-)

model/protocol.cto
martinhalford: We could do but it felt like overkill for a single data field - it's just a data bucket at the moment.

If this needs to become more complicated, say, with multiple named fields and data types then we should def. make it a concept.

model/protocol.cto
martinhalford: Have standardised on camel case.

PR #31, opened 2025-03-02
Templates and agreement partial fetch and pagination #15

### LINKED ISSUE #15: templates and agreement partial fetch and pagination
The fetch of a list of templates returns a list of all templates, this could lead to a very large response. Is there an alternative pattern we could use where the list of template names is fetched and then individual fetch template requests can be made? I am worried about the size of the response for large templates and unnecessary response overhead. Another possibility would be a param supplied to the request to indicate if a full list of templates would be returned or just a list of ids. If a full list is returned maybe there could be a max size/number of responses returned or it maybe it could be paginated. Perhaps a list of just the template metadata is a good compromise too.

Marvinrose:
Fixes #15 by addressing large response concerns for GET /templates and GET /agreements:
- Added Pagination: Both endpoints support offset and limit (capped at 100 server-side), with meta (total, offset, limit) and links (self, next, prev, first, last) in the response

- Implemented `fields` param for Partial Fetching: The fields query param (e.g., ?fields=id,name) allows clients to fetch specific fields, supporting the alternative pattern of listing IDs/names followed by individual requests (e.g., GET /templates/{name}).

- Added `full` param to toggle full objects (e.g., `?full=true`).

- Updated `openapi.json` to document these params, removed `limit` maximum to allow server-side capping.

- Tested with Postman to confirm functionality.

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -32,11 +32,7 @@
             "type": "string"
           }
         },
-        "required": [
-          "$class",
-          "base64",
-          "mimeType"
-        ],
+        "required": ["$class", "base64", "mimeType"],
         "$decorators": {
           "description": [
             "Bytes and mime type for a blob of data, such as images, files, etc."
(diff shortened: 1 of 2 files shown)


dselman:
This has been implemented in the generic REST handler.

PR #43, opened 2025-03-17
Add Standard 'appearance' Property for Better Browser Compatibility

moodyadi2006:
This pull request adds the standard `appearance` property alongside `-webkit-appearance` to address a warning in modern browsers. This change enhances compatibility and prevents potential rendering issues, ensuring a smoother user experience across different browsers.


dselman:
I don't think this css comes from this project. The fix should go into the upstream slate project.

Issue #52, opened 2025-04-05
adding a sample request structure to Template


niallroche:
CTA files contain a sample request json that can be useful for testing the logic for a clause/contract. Multiple requests may be needed when there are multiple functions defined. A suggestion is to align the APAP Template object closer to a CTA and to add an optional request attribute to the Template.

These examples are useful for implementers to have a better idea of the request structure to use but could also be useful as a way to verify the operation of a template when created. This could form part of a unit test when a template is uploaded, particularly if code generation is run on the server side and would add as a useful check to ensure that the code logic is operating properly

Marvinrose:
Hi @niallroche 

I sent a pr that addresses this issue, kindle review please.

PR #55, opened 2025-04-08
Add optional request attribute to Template (#52)

Marvinrose:
Added an optional request attribute to org.accordproject.protocol@1.0.0.Template in openapi.json, updated server implementation and documented in specification.md. Successfully tested with a sample request, aligning APAP with CTA files for testing template logic.

Screenshot (583)

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -32,11 +32,7 @@
             "type": "string"
           }
         },
-        "required": [
-          "$class",
-          "base64",
-          "mimeType"
-        ],
+        "required": ["$class", "base64", "mimeType"],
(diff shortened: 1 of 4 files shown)


Marvinrose:
Hi @niallroche 

Kindly review please.

dselman:
Changes to the model should be made to protocol.cto and then we can use codegen to create TS files (either from the Concerto model itself using Concerto Codegen or via generation of the OpenAPI and then OpenAPI codegen).

https://concerto.accordproject.org/docs/category/code-generation

PR #65, opened 2025-06-22
Added Template Keyword Search (#46)

### LINKED ISSUE #46: Add Template Keyword Search via GET /templates?keywords
The `GET /templates` endpoint supports pagination and partial fetching, but lacks a way to search templates by their `keywords` field.  Adding a `keywords` query param (e.g., `?keywords=contract,legal`) would enable efficient server-side filtering.

### Proposed Solution
- Update `GET /templates` in `openapi.json` with an optional `keywords` param.
- Param: `?keywords=tag1,tag2` (comma-separated, matches any keyword in `Template.keywords`).
- Response: Paginated array of matching `Template` objects..
- Example: `GET /templates?keywords=contract,legal&limit=10` returns templates tagged with "contract" or "legal".
- Document in `specification.md` with examples.

### Benefits
- Reduces client-side filtering overhead for large template sets.
- Leverages existing `keywords` field in `Template` schema.

Marvinrose:
Resolves #46: Add Template Keyword Search via GET /templates?keywords
- Added keywords filtering in server/crud.ts using arrayFilters and PostgreSQL && operator.
- Updated server/templates.ts to use enhanced defaultWhereClause.
- Exported db in server/db/index.ts for seeding.
- Implemented seeding in server/db/seeders/seed.ts using raw SQL due to persistent TypeScript error.
- Updated openapi.json with keywords parameter and paginated response.
- Documented in specification.md with example.

Screenshot (591)
Screenshot (590)
Screenshot (589)
.

### DIFF
--- a/client/typescript/apap.ts
+++ b/client/typescript/apap.ts
@@ -7,138 +7,138 @@
 export interface paths {
   "/sharedmodels": {
     /**
-     * List All Sharedmodels 
+     * List All Sharedmodels
      * @description Gets a list of all `sharedmodel` entities.
      */
     get: operations["listSharedmodels"];
     /**
-     * Create a Sharedmodel 
+     * Create a Sharedmodel
      * @description Creates a new instance of a `sharedmodel`.
      */
     post: operations["createSharedmodel"];
   };
   "/sharedmodels/{uri}": {
     /**
-     * Get a sharedmodel 
+     * Get a sharedmodel
      * @description Gets the details of a single instance of a `sharedmodel`.
      */
     get: operations["getSharedmodel"];
     /**
-     * Update a sharedmodel 
+     * Update a sharedmodel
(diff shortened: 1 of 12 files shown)


dselman (changes requested):
You seem to have changed single quotes to double quotes which is introducing a lot of noise into the diff and making it unnecessarily hard to review. Please check your diff and make the minimal changes. Do not leave commented out code. Beware of files that are generated (like openapi.json) - you should not change those.

The approach should not be to special case "keywords" but to implement this in a generic way so that it works for any string[], including those that might be added in the future.

In default filter you can see the todo comment about only supporting string lhs and rhs, that is where you should add the check for string[].

Marvinrose:
Hi @dselman

dselman (changes requested):
Getting there! Still some rough edges.

Issue #63, opened 2025-06-18
Add names for CTO Files


niallroche:
# Feature Request 🛍️

Add the ability to specify a filename for each CTO uploaded in a org.accordproject.protocol@1.0.0.CtoModel

## Use Case

The API should have the ability to specify a filename for each CTO File uploaded to be consistent with how files are managed locally at edit time and for runtime processing of triggers

## Possible Solution

Add an array of filenames to the CTO inputs

## Context

Without proper file names, an arbitrary filename would need to be applied such as model1.cto, model2.cto etc. however this may impact on imports between model files that expect a particular filename

dselman:
I've implemented this via a change to protocol.cto.

PR #66, opened 2025-06-30
Added filenames support for CTO file uploads in CtoModel (#63)

Marvinrose:
This pull request addresses issue #63 by adding support for the optional filenames property in the 
org.accordproject.protocol@1.0.0.CtoModel schema for the /sharedmodels endpoint. 

The changes ensure that the endpoint correctly validates and stores SharedModel instances with filenames, handles namespace conflicts, and returns responses

Fixes:

- Added o String[] filenames optional to the CtoModel concept.
- Regenerated the base64-encoded MODEL constant in schema.ts.
- Enhanced validation to check filenames length matches ctoFiles length in concertovalidation.ts.
- Generated openapi.json

Screenshot (592)


niallroche:
there is a potential security issue here with the structure of filenames for the CTO files if they will be used to create a .cta file. That might be out of scope for a reference implementation and API definition, but something to bear in mind. A reasonable set of checks for platform agnostic valid filenames could possibly be considered (e.g must contain once valid character and end in .cto as a starting point.

dselman (changes requested):
I think it would be cleaner to make a breaking change to `protocol.cto` and introduce a new concept, something like:

This will remove the need for the special case validation.

```
concept CtoFile {
   o String contents
   o String filename optional // required?
}
```

PR #75, opened 2025-11-09
feat(agreements): add trigger init, payload validation and state persistence

mttrbrts:


server/package.json
```diff
@@ -49,11 +49,13 @@
     "@types/jest": "^29.2.5",
     "@types/morgan": "^1.7.35",
     "@types/node": "^10.17.60",
+    "@types/supertest": "^6.0.3",
     "axios": "^1.6.0",
     "concurrently": "^6.2.0",
     "drizzle-kit": "^0.31.0",
     "jest": "^29.7.0",
     "nodemon": "^1.18.10",
+    "supertest": "^7.1.4",
```
dselman: dev dependency, no?

server/scripts/export-template.js
```diff
@@ -0,0 +1,237 @@
...
+  const accordproject = pkg.accordproject || {};
+
+  // Read model (CTO) files
+  const modelDir = path.join(templateDir, 'model');
+  let ctoFiles = [];
+  let mainModelContent = null;
+  let namespace = null;
+  
+  if (fs.existsSync(modelDir)) {
+    const files = fs.readdirSync(modelDir)
```
dselman: Load the dir using TemplateArchive.fromDirectory and then use the API get the ASTs / logic etc and build the model for APAP?

dselman (approved):
Some nits but looks great overall.

server/test/archives/latedeliveryandpenalty-typescript/logic/logic.ts
```diff
@@ -0,0 +1,140 @@
...
+                count: 0,
+                lateDeliveryProcessed: false,
+                totalPenalties: 0.0,
+            }
+        }
+    }
+
+    /**
+     * Convert Duration to milliseconds
+     * Manual conversion without requiring dayjs duration plugin, note that this is simplified and will likely break with daylight savings time changes and leap years
```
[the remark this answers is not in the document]
mttrbrts: Yes, we should. Let's do that separately

Issue #64, opened 2025-06-18
URI handling for templates


niallroche:
# Discussion 🗣
Templates have Uris but also ids to identify them, this can cause some issues.

Templates have a URI but also a separate id field that is assigned by the server when templates are created. This separation is useful as, depending on the storage mechanism used by the APAP server, URIs may not be valid identifiers for the storage mechanism used (e.g. filename/database field). 

We have the issue around potential conflicts where the same uri scheme could be used for different templates, and these would have different ids and the same uri and we need to define the appropriate behaviour for managing this.

In the current implementation agreements refer to templates by URI, not id so there is an issue here if more than one template shares the uri. Therefore if this is the desired behaviour, then a validation rule will need to be added where a uri is checked to be unique when creating templates and a 409 Conflict response returned.

Agreements also have URIs, so presumably similar consistant logic will need to apply here too.

There is a related concern where a user tries to create an agreement that refers to a valid uri, but one that they did not create. Maybe we need a public flag for some templates? I am conscious that this can get tricky with access control groups and template/agreement sharing within an organisational group.

As mentioned on the working group call, thoughts are welcome @dselman

PR #87, opened 2026-01-29
feat: return 409 Conflict on unique constraint violation

rockaxorb13:
### Description
This PR addresses the handling of duplicate resource identifiers (URIs) for Templates and Agreements.

Previously, attempting to create a resource with a `uri` that already existed would trigger a database-level constraint violation, resulting in an unhandled `500 Internal Server Error`. This behavior was noted in the discussion of Issue #64 regarding behavior for managing unique identifiers.

### Changes
* **Modified `server/handlers/crud.ts`**: Updated the `POST` and `PUT` error handlers to explicitly catch PostgreSQL error code `23505` (`unique_violation`).
* **Response Handling**: The server now returns a standard `409 Conflict` HTTP status code with a descriptive error message when a uniqueness constraint is violated, rather than a generic server error.

### Related Issues
* Addresses parts of #64 (Handling conflicts for URI uniqueness).

### Verification
* Ran `npm test` in `server/` directory.
* **Result**: All 37 tests passed, confirming no regression in existing CRUD logic.

### DIFF
--- a/server/handlers/crud.ts
+++ b/server/handlers/crud.ts
@@ -266,7 +266,15 @@ export function buildCrudRouter<T extends PgTable<any> & TableWithId>({
                 }
 
                 res.json(inserted[0]);
-            } catch (error) {
+            } catch (error: any) {
+                // Handle Unique Constraint Violation (Duplicate URI)
+                if (error?.code === '23505') {
+                    return res.status(409).json({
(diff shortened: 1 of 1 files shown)


rockaxorb13:
Be gone, bot.

PR #90, opened 2026-02-01
chore: fix CI workflow swallowing errors & add timeout

rockaxorb13:
Previously, the build.yml file was not displaying the build errors just giving a clean-slate "tests not configured" output. This PR changes that, runs the tests in the server directory to test the API code and run the logic checks defined there, as 'Jest' is configured in the `server/` directory. 

This PR also removes the matrix system of testing for windows and mac as almost all servers use linux, however if need be, I can configure it to test all 3 operating systems as well. 

Additionally integrated a 60s timeout flag to prevent the network errors I found when testing locally.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,7 +7,7 @@ on:
   pull_request:
     branches:
       - main
-      - v4.0.0 # Temporary, while we prepare for merge to main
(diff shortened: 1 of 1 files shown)


rockaxorb13:
Be gone, Bot.

PR #129, opened 2026-03-19
fix: prevent ```Agreement.uri``` from overriding resource URI

### LINKED ISSUE #128: ```Agreement.uri``` field overrides the resource URI in ```getAgreements```
Hey, ran into annoying bug in getAgreements.
When mapping over agreements.items, we spread ...a at the end of the returned object, which means any field on the Agreement schema silently overrides what we set explicitly, including uri. So the apap://agreements/${a.id} we carefully construct gets thrown away if Agreement happens to have a uri field on it.
```
contents: agreements.items.map((a: typeof Agreement) => {
    return {
        uri: `apap://agreements/${a.id}`,
        mimeType: "application/json",
        text: JSON.stringify({ ...a.data, $identifier: a.id }, null, 2),
        ...a
    }
})
```
and we can see the Agreement schema has ```uri``` attribute:
```
export const Agreement = pgTable("Agreement", {
   id: serial().primaryKey(),
   uri: text().unique().notNull(),
   data: json().notNull(),
   template: text(),
   state: json(),
   agreementStatus: AgreementStatusType().notNull(),
   agreementParties: json().array(),
   signatures: json().array(),
   historyEntries: json().array(),
   attachments: json().array(),
   references: text().array(),
   metadata: json(),
});
```

The result is: 

Image

We can see the unexpected result: ```uri: "apap://my-test-agreement-1"```

Thomas-Sedhom:
# Problem
 In ```getAgreements```, the object spread ...a was placed after the explicitly defined uri field. Since Agreement has its own uri property, it was silently overwriting the apap://agreements/${a.id} URI we were trying to set — meaning MCP clients were getting the wrong URI back.

# Solution
Moved ```...a``` to the top of the returned object so that our explicit ```uri```, ```mimeType```, and text fields always take precedence over whatever's on the schema.

```
// before
return {
    uri: `apap://agreements/${a.id}`,   // overridden by ...a
    mimeType: "application/json",
    text: JSON.stringify({ ...a.data, $identifier: a.id }, null, 2),
    ...a
}

// after
return {
    ...a,
    uri: `apap://agreements/${a.id}`, 
    mimeType: "application/json",
    text: JSON.stringify({ ...a.data, $identifier: a.id }, null, 2),
}
```
And it finally return the correct URI: 

image

Fixes: #128

### DIFF
--- a/server/handlers/mcp.ts
+++ b/server/handlers/mcp.ts
@@ -62,6 +62,7 @@ async function getTemplates(uri: URL) {
         return {
             contents: templates.items.map((t: typeof Template) => {
                 return {
+                    ...t,
                     uri: `apap://templates/${t.id}`,
                     mimeType: "application/json",
                     text: JSON.stringify(t)
@@ -84,10 +85,10 @@ async function getAgreements(uri: URL) {
         return {
             contents: agreements.items.map((a: typeof Agreement) => {
                 return {
+                    ...a,
                     uri: `apap://agreements/${a.id}`,
                     mimeType: "application/json",
                     text: JSON.stringify({ ...a.data, $identifier: a.id }, null, 2),
-                    ...a
                 }
             })
         }
--- a/server/package-lock.json
+++ b/server/package-lock.json
@@ -8145,6 +8145,27 @@
         "node": ">=8"
       }
     },
+    "node_modules/escodegen": {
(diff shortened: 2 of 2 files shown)


Thomas-Sedhom:
This work has been continued in #164

PR #133, opened 2026-03-19
test: add unit tests for ```CRUD``` router

Thomas-Sedhom:
Added a full unit test suite for the buildCrudRouter utility in crud.ts. The tests cover all five HTTP methods the router exposes and use a mock DB layer so there's no real database dependency.
What's covered

GET /

- Returns paginated results and correctly caps limit at 100 even when a higher value is passed
- Enriches user items with an empty roles array when typeName is users

POST /

- Runs schema validation, custom validation, transformRequest, and transformResponse in the right order and with the right data
- Returns 400 with field-level details when schema validation fails
- Returns 400 with custom error messages when custom validation fails
- Returns 409 on duplicate key constraint violations (pg error code 23505)

GET /:id

- Returns a single item passed through transformResponse
- Returns 404 when the record doesn't exist
- 

PUT /:id

- Updates the record and returns it transformed
- Returns 409 on unique constraint violation

DELETE /:id

- Deletes the record and returns { status: 'deleted' }

image
image

### DIFF
--- a/server/handlers/crud.test.ts
+++ b/server/handlers/crud.test.ts
@@ -1,0 +1,448 @@
+import request from 'supertest';
+import express from 'express';
+import { jest } from '@jest/globals';
+import { z } from 'zod';
+import { SharedModel } from '../db/schema';
+import { buildCrudRouter } from './crud';
+
+type QueryChainOptions = {
+    result?: any;
+    error?: any;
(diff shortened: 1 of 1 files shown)


Thomas-Sedhom:
This work has been continued in #165

PR #154, opened 2026-03-28
fix: check concerto validation result before calling templateFromDatabase

### LINKED ISSUE #138: Bug: templateValidation calls templateFromDatabase before concerto validation success is checked
# Bug Report 🐛

The template validation flow in `server/handlers/templates.ts` currently calls `templateFromDatabase(body)` before checking whether the initial Concerto validation actually succeeded.

## Expected Behavior

If `concertoValidation('Template', body)` fails, the function should return immediately and avoid running template build logic.

## Current Behavior

Inside `templateValidation(body)`:

1. `concertoValidation('Template', body)` is called
2. `templateFromDatabase(body)` is called immediately after
3. only then does the code check `if (!result.success)`

This means invalid template payloads can still trigger template build work even when the request has already failed validation.

## Steps to Reproduce

1. Send an invalid template payload into the template validation path
2. Observe that `templateFromDatabase(body)` is still reached before the validation result is checked
3. The function may then surface follow-on build errors instead of stopping cleanly at the original validation failure

## Why This Matters

This causes unnecessary work on invalid input and can make error reporting less clear by allowing later build failures to mask the original validation failure.

## Possible Solution

Reorder the function so that:
- Concerto validation runs first
- failed validation returns immediately
- `templateFromDatabase(body)` runs only after validation succeeds

## Context

I found this while reviewing the APAP template validation flow locally. The issue appears to be a small control-flow bug, and I have already prepared a minimal fix for it.

JayDS22:
## Summary

Fixes #138

## Problem

In `server/handlers/templates.ts`, the `templateValidation` function called
`templateFromDatabase(body)` before checking whether `concertoValidation`
returned success or failure. When the body was invalid, `templateFromDatabase`
threw a deserialization error that got caught by the outer catch block,
masking the structured validation error that concertoValidation had already
produced.

## Fix

Moved `templateFromDatabase` to after the `result.success` check. Invalid
payloads now return the Concerto validation error directly. Template
deserialization only runs on validated data.

## How to test

1. POST to `/templates` with a body missing required fields
2. Before: 500 with a deserialization error
3. After: 400 with structured validation details

Discord: Jay Guwalani (jayds22)

### DIFF
--- a/server/handlers/templates.ts
+++ b/server/handlers/templates.ts
@@ -9,10 +9,15 @@ const router = express.Router();
 async function templateValidation(body:any) : Promise<ValidationResult> {
     try {
         const result = await concertoValidation('Template', body);
-        const template = await templateFromDatabase(body);
+
+        // Return validation errors before attempting deserialization,
+        // otherwise templateFromDatabase can throw and mask the real failure.
         if(!result.success) {
             return result;
         }
+
+        const template = await templateFromDatabase(body);
+
         return {
             success: true,
             data: result.data


JayDS22:
@github-actions unstale

JayDS22:
@github-actions unstale

PR #153, opened 2026-03-28
fix(mcp): surface HTTP status and API error details in MCP tool error responses

JayDS22:
Closes #152

### Problem

All MCP tools and resource handlers in `handlers/mcp.ts` discard the HTTP status code and error body when `makeApiRequest()` returns a non-ok response. Every failure produces a generic `throw new Error('Failed to load ...')` regardless of whether the REST API returned 400, 404, or 500. MCP clients receive identical error messages for completely different failure modes.

### Changes

**Added `buildApiErrorMessage()` helper** that extracts the HTTP status code and response body from failed API responses, producing errors like:
Failed to load template 'abc-123' (HTTP 404): {"error": "Template not found"}

instead of:
Failed to load template

**Updated all 8 error paths** in `mcp.ts`:
- `getAgreement()` - now includes agreement ID and HTTP status
- `getTemplates()` - now includes HTTP status and API error body
- `getAgreements()` - now includes HTTP status and API error body
- `draftAgreement()` - now includes agreement ID, format, and HTTP status
- `triggerAgreement()` - now includes agreement ID and HTTP status
- Template resource handler (single fetch) - now includes template ID and HTTP status
- `getTemplate` tool - now includes template ID and HTTP status
- `getAgreement` tool - now includes agreement ID and HTTP status

Also improved error logging in the two resource list handlers (agreements and templates) which already returned empty arrays on failure but were logging without status context.

### Why This Matters

MCP clients like Claude Desktop and ChatGPT need to distinguish between "this template doesn't exist" (404) and "the server is down" (500) to provide useful feedback to users. This change makes every MCP error actionable.

PR #155, opened 2026-03-28
fix: prevent Agreement.uri from overriding MCP resource URI in getAgreements

### LINKED ISSUE #128: ```Agreement.uri``` field overrides the resource URI in ```getAgreements```

JayDS22:
## Summary

Fixes #128

## Problem

In `server/handlers/mcp.ts`, the `getAgreements` function built MCP content
objects by setting `uri` first, then spreading `...a` (the full Agreement
database row) last. The Agreement row has its own `uri` field, so the spread
overwrote the MCP resource URI (`apap://agreements/{id}`) with the agreement's
data URI (`resource:org.accordproject...`). MCP clients could not resolve the
resource because they received the wrong URI scheme.

Additionally, spreading the entire row added unrelated database fields
(`template`, `state`, `agreementStatus`, etc.) to the content object, which
only expects `{ uri, mimeType, text }`.

## Fix

Removed the `...a` spread from the contents mapping. The agreement data is
already serialized inside the `text` property. Placed `uri` last in the
object literal so the MCP resource URI always takes precedence.

## How to test

1. Create an agreement in the database
2. Connect an MCP client and read `apap://agreements`
3. Before: content items have `uri` like `resource:org.accordproject...`
4. After: content items have `uri` like `apap://agreements/1`

Discord: jayds22

### DIFF
--- a/server/handlers/mcp.ts
+++ b/server/handlers/mcp.ts
@@ -32,6 +32,14 @@ async function makeApiRequest(url: string, options: RequestInit = {}) {
     });
 }
 
+// Builds a meaningful error message from a failed API response so that MCP clients
+// can tell apart a 404 (resource missing) from a 400 (bad input) or a 500 (server issue).
+// Without this, every failure just says "Failed to load ..." which is impossible to debug.
+async function buildApiErrorMessage(result: globalThis.Response, context: string): Promise<string> {
+    const body = await result.text().catch(() => 'No error details available');
+    return `${context} (HTTP ${result.status}): ${body}`;
+}
+
 async function getAgreement(uri: string, { agreementId }: { agreementId: string }) {
     console.log(`Fetching agreement with ID: ${agreementId}`);
     const url = new URL(uri);
@@ -48,8 +56,10 @@ async function getAgreement(uri: string, { agreementId }: { agreementId: string 
         };
     }
     else {
-        console.error(`Failed to load agreement with ID: ${agreementId}`);
-        throw new Error('Failed to load agreement');
+        // Surface the actual status code and API response so the client knows what went wrong
+        const errorMsg = await buildApiErrorMessage(result, `Failed to load agreement '${agreementId}'`);
+        console.error(errorMsg);
+        throw new Error(errorMsg);
     }
 }
 
@@ -70,8 +80,10 @@ async function getTemplates(uri: URL) {
         }
     }
     else {
-        console.error('Failed to load templates');
-        throw new Error('Failed to load template');
+        // Previously just threw "Failed to load template" with no status or details
+        const errorMsg = await buildApiErrorMessage(result, 'Failed to load templates');
+        console.error(errorMsg);
+        throw new Error(errorMsg);
     }
 }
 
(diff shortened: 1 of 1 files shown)


JayDS22:
@github-actions unstale

JayDS22:
@github-actions unstale

niallroche:
this cannot merge with 153, I bring to bring the test file from #153's branch into this one

JayDS22:
Sounds good. Happy to do the lift: I'll cherry-pick `mcp.test.ts` from #153's branch onto this one, add export to `buildApiErrorMessage` so the test imports it instead of redeclaring, and close #153 once this merges. Want me to push to this branch directly or open the change as a follow-up commit for your review?

JayDS22:
Force-pushed a rebase onto current main (d9168ba → e13b77f). The branch was stale (based on af4c9a0, pre-#154/#171/#176), so the rebase also dropped unintended drift in templates.ts,  workflow YAML, and lockfiles, that also explains the previous Windows 20.x failure (was npm ci failing on a stale lockfile in ~12s).                                                  Diff is now purely server/handlers/mcp.ts (+56/−14). Local jest green (39/39).                                                                                                           

Heads-up on the new Windows red check, it's unrelated to this PR and pre-existing on main. The failure is in root npm run build at npx reslate build:                                   
```
Error: spawn EINVAL                                                                                                                                                                      
at ChildProcess.spawn (node:internal/child_process:414:11)
at Object.<anonymous> (D:\a\apap\apap\node_modules\reslate\reslate.js:21:6)          
```                                                                                                  
                
Classic Windows-Node spawn gotcha in reslate.js. Same job is red on main's current HEAD (95105f5, run 26512754891) and on the two prior main runs back to early May; Linux + macOS green on all of them. Happy to open a separate PR for it.
                                                                                                                                                                                           
@niallroche @mttrbrts - could one of you approve the Build / Unit Tests matrix when you get a chance? Thanks.

mttrbrts (approved):

PR #163, opened 2026-04-14
fix(bug): number id parse issue

hemantch01:
## Description
Fixes #162

this pr resolves a vulnerability where numeric `:id` route parameters were partially parsed due to the use of `parseInt()`. For example, requests like `GET /templates/1abc` were incorrectly truncated and executed as `GET /templates/1`, potentially leading to unexpected resource modification or deletion.
### Tests passed
Screenshot 2026-04-14 100207

### Postman screenshot after FIX
Screenshot 2026-04-14 101328

PR #169, opened 2026-04-18
fix: sanitize pagination params to prevent NaN in sql queries 

### LINKED ISSUE #168: bug: invalid pagination params cause `NaN` to leak into database queries
## expected behavior
when someone passes non-numeric values like `page=abc` or `limit=foo` in query params, the server should either return a 400 error or quietly fall back to safe defaults (page 1, limit 100).

## current behavior
`parseInt("abc")` returns `NaN`, and `Math.max(1, NaN)` is still `NaN`. this NaN gets passed straight into drizzle's `.limit()` and `.offset()`, which produces broken sql or weird postgres errors instead of a clean response.

## steps to reproduce
1. start the server locally
2. send a request like `GET /agreements?page=abc&limit=foo`
3. observe the server throws an internal error or returns malformed results
4. check the logs — you'll see NaN flowing into the sql query params

## context
found this while testing the crud pagination in `crud.ts` . the `parseQueryParams` function doesn't guard against non-numeric input at all. any route using `buildCrudRouter` (agreements, templates, shared models) is affected.

### desktop
- os: linux
- browser: was using postman
- version: current main branch

hemantch01:
# fixes #168

when non-numeric values like `page=abc` or `limit=foo` are passed as query params,
`parseInt` returns `NaN` which flows straight into drizzle's `.limit()` and `.offset()`,
producing broken sql or postgres errors.

added `Number.isNaN()` checks in `parseQueryParams` so invalid input falls back
to safe defaults (page 1, limit 100) instead of crashing.

affects all routes using `buildCrudRouter` — agreements, templates, shared models.

### DIFF
--- a/server/handlers/crud.ts
+++ b/server/handlers/crud.ts
@@ -137,9 +137,12 @@ function parseQueryParams(req: Request): QueryParams {
         ...filters
     } = req.query;
 
+    const parsedPage = parseInt(page as string);
+    const parsedLimit = parseInt(limit as string);
+
     return {
-        page: Math.max(1, parseInt(page as string)),
-        limit: Math.min(100, Math.max(1, parseInt(limit as string))),
+        page: Math.max(1, Number.isNaN(parsedPage) ? 1 : parsedPage),
+        limit: Math.min(100, Math.max(1, Number.isNaN(parsedLimit) ? 100 : parsedLimit)),
         sortBy: sortBy as string,
         sortOrder: (sortOrder as string).toLowerCase() === 'desc' ? 'desc' : 'asc',
         filters


Rishabh060105:
Hi @hemantch01, @mttrbrts I have pushed a follow-up to #169 to further strengthen the pagination fix related to #168.

While the original change correctly handled the `NaN` case from `parseInt`, it could still permit partially numeric or non-decimal inputs such as `12abc` or `0x10` to resolve into unintended pagination values. This follow-up tightens the parsing so that `page` and `limit` only accept whole decimal strings, continue to enforce the existing `limit <= 100` cap, and fall back to safe defaults when malformed values are supplied.

I have also added focused regression coverage for invalid inputs and repeated query parameters to ensure this behavior remains stable going forward.

For validation, I ran the full relevant local checks successfully:
- `npm run build` in `server`
- `npm test` in `server`
- `npm test` at the repository root

All local unit tests are passing: 4 test suites, 42 tests.

hemantch01:
hi @Rishabh060105 pr has already been raised avoid duplicate work.

niallroche:
Same situation as #163 — the NaN guard here is correct but #177 covers the same code path with a stricter validator that also handles hex, partial strings, and repeated query params. Closing in favour of [#177](https://github.com/accordproject/apap/issues/177). The issue report ([#168](https://github.com/accordproject/apap/issues/168)) was the right call; the fix just landed in a more complete form elsewhere. Thanks again @hemantch01 apologies we had a backlog for a while and some duplicate effort. We are aiming to clean up to allow more PRs

PR #178, opened 2026-05-28
test(templates): add validation order tests for issue #138

### LINKED ISSUE #138: Bug: templateValidation calls templateFromDatabase before concerto validation success is checked

Satvik77777:
## Summary
Adds test coverage for the validation order fix in #154.

## What this PR does
- Tests that invalid body returns 400 with Concerto error (not 500)
- Tests that valid body proceeds correctly to templateFromDatabase
- Tests that DB errors return 400, not masked as 500

## Key finding
jest.mock('../db/schema') auto-mocks TemplateInsertSchema making 
safeParse undefined. Tests explicitly mock safeParse per case.

## References
- Closes #138
- Builds on fix by @JayDS22 in #154

## Author Checklist
- [x] DCO sign-off provided
- Tests added and passing (5/5)


Satvik77777:
Hi @niallroche — you recently merged #154 which fixed the 
validation order bug in templateValidation. This PR adds the 
missing test coverage for that fix. All 4 tests pass and DCO 
is signed off. Would appreciate a review when you get a chance!

JayDS22:
Thanks for the additional test coverage on the validation flow.                                                                                                                            
                                                                                                                                                                                             
  One substantive note: as written, the four tests don't actually catch a regression of the #154 fix. The bug was that `templateFromDatabase` ran *before* the `result.success` check, so its
   throw masked the real Concerto error. None of the tests exercise that ordering; they'd all pass against the buggy pre-#154 code too. To lock the fix in:                                  
                                                                                                                                                                                             
  ```ts                                                     
  it('returns concerto error, not DB error, when both would fail', async () => {
    (mockedSchema.safeParse as any) = jest.fn().mockReturnValue({ success: true, data: {} });
    require('./concertovalidation').concertoValidation.mockResolvedValueOnce({                                                                                                               
      success: false, error: { errors: [{ message: 'Concerto error' }] }
    });                                                                                                                                                                                      
    const dbSpy = jest.spyOn(db, 'templateFromDatabase')    
      .mockRejectedValueOnce(new Error('should not be called'));                                                                                                                             
                                                                                                                                                                                             
    const res = await request(app).post('/templates').send({});
    expect(res.body.details[0].message).toBe('Concerto error');                                                                                                                              
    expect(dbSpy).not.toHaveBeenCalled();   // load-bearing assertion
  });                                                                                                                                                                                        
  ```
                                                                                                                                                                                             
  The `not.toHaveBeenCalled()` is what makes a future revert of #154 fail this suite. Without it, the order is unconstrained.                                                                
   
  Nit: `jest.spyOn(TemplateInsertSchema, 'safeParse').mockReturnValue(...)` reads cleaner than reassigning `safeParse` per test, and auto-restores between cases.

Satvik77777:
Thanks @JayDS22 — that's a really good point, I hadn't thought about it that way. You're right that the original tests would pass even against the buggy pre-#154 code since they don't actually enforce the ordering.
I've added the missing test in commit 742b178 — the not.toHaveBeenCalled() assertion is the key part that makes a future regression actually fail the suite. All 5 tests passing now locally and on CI.
Appreciate you taking the time to review so carefully

dselman (approved):

PR #179, opened 2026-05-28
test(crud): add PUT validation tests for issue #136

### LINKED ISSUE #136: Bug: PUT routes in buildCrudRouter bypass schema and custom validation
# Bug Report 🐛

`PUT /:id` updates in `buildCrudRouter` currently bypass the validation flow that is already enforced for `POST /`.

## Expected Behavior

If a router is configured with `validateBody.schema` and/or `validateBody.custom`, both `POST` and `PUT` requests should run the same validation checks before persisting data.

## Current Behavior

In `server/handlers/crud.ts`:

- `POST /` validates request bodies using the configured schema and custom validator
- `PUT /:id` updates the resource without running that validation path

This means invalid update payloads may be accepted even when the same payload would be rejected during creation.

## Steps to Reproduce

1. Configure a router with `validateBody.schema` and/or `validateBody.custom`
2. Send an invalid payload to `POST /`
3. Observe that the request is rejected
4. Send the same invalid payload to `PUT /:id`
5. Observe that the update path does not apply the same validation flow

## Why This Matters

This creates inconsistent API behavior and allows invalid data to bypass validation through updates.

## Possible Solution

In the `PUT /:id` handler inside `buildCrudRouter`, apply the same validation flow already used in the `POST /` handler:
- run `validateBody.schema.safeParse(req.body)`
- run `validateBody.custom(req.body)` when configured
- return `400` with validation details if validation fails

## Context

I found this while reviewing the generic CRUD router in APAP. The create path already validates correctly, so this looks like a missing validation step in the update path rather than a broader design issue.

Satvik77777:
## Summary
Fixes the PUT validation bug from #136 and adds full test coverage.

### Fixes (addressing Copilot review on #137)
- Custom validator now runs independently of schema validation
  (previously only ran inside the schema branch — so custom-only
  configs were silently skipped)
- Added empty body guard to PUT handler returning 400
- Added transformRequest support to PUT handler for parity with POST

### Tests (6 passing)
- Rejects invalid PUT body with 400 on schema failure
- PUT and POST apply identical validation behavior  
- Accepts valid PUT body and updates resource
- Rejects PUT body that fails custom Concerto validation
- Rejects empty PUT body with 400
- Runs custom validation even when no schema is configured

## References
- Closes #136
- Supersedes closed PR #137 by @Ashvin-KS
- Addresses Copilot review feedback from #137

## Author Checklist
- [x] DCO sign-off provided
- [x] Bug fix included
- [x] Tests added and passing (6/6)

### DIFF
--- a/server/handlers/agreements.ts
+++ b/server/handlers/agreements.ts
@@ -8,6 +8,14 @@ import { TemplateArchiveProcessor } from '@accordproject/template-engine';
 import { HttpTemplateRetriever } from './retrievers/HttpTemplateRetriever';
 import { Template as CiceroTemplate } from '@accordproject/cicero-core';
 
+/**
+ * @param db The database handle stored on `res.locals.db`.
+ * @param agreementId The numeric agreement identifier from the request path.
+ * @return The agreement row, its associated template row, and the reconstructed Cicero template.
+ * @details Loads an agreement from the database, resolves its backing template either
+ * through the cached `templateHash` or the stored template URI, and rebuilds the
+ * runtime template archive needed for draft and trigger operations.
+ */
 async function resolveAgreement(db: any, agreementId: string) {
     console.log('Getting agreement: ' + agreementId);
     const result = await db.select().from(Agreement).where(eq(Agreement.id, Number.parseInt(agreementId))).limit(1);
@@ -47,6 +55,14 @@ async function resolveAgreement(db: any, agreementId: string) {
 
 const router = express.Router();
 
+/**
+ * @param req The Express request containing the agreement payload to create.
+ * @param res The Express response used to return the created agreement or an error.
+ * @return Resolves after the agreement creation response has been written.
+ * @details Validates the incoming agreement body with Zod and Concerto, optionally
+ * resolves and caches a remote template archive when the template URI matches a supported
+ * retriever, and finally inserts the agreement into the database.
+ */
 router.post('/', async (req, res) => {
     try {
         const db = res.locals.db;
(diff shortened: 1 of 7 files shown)


Satvik77777:
Hi @niallroche — this PR fixes #136 which was left open after 
#137 was auto-closed. It also addresses the two issues flagged 
by Copilot in #137's review: custom validation running only 
inside the schema branch, and missing empty body guard. 
6 tests passing, DCO signed

JayDS22:
Nice catch on the schema/custom independence. That's the real issue Copilot was pointing at on #137. The empty-body guard and POST/PUT symmetry test are both solid additions.             
                                                                                                                                                                                             
  One concern with **test 6** ("runs custom validation even when no schema is configured"): I think it passes for the wrong reason. `jest.mock('../db/schema')` auto-mocks                   
  `TemplateInsertSchema.safeParse` to return `undefined`, so `result.success` is falsy and the **schema branch** returns 400 first, before the custom branch ever runs. The test still hits  
  400, but not via the path its name describes, so it doesn't actually prove that schema-independent custom validation works.                                                                
                                                            
  To exercise that path, you'd want a router built directly with `buildCrudRouter({ ..., validateBody: { custom: ... } })` (no `schema` key), since `templatesRouter` always has a schema    
  configured.
                                                                                                                                                                                             
  Also worth surfacing in the PR description: the diff adds `if (transformRequest) { req.body = transformRequest(req); }` to the PUT handler. If POST already has it (parity), totally fine, 
  but it's outside the validation scope the description focuses on, easy to miss in review.

Satvik77777:
Thanks @JayDS22 — you're right on test 6. The mock was 
intercepting at the schema branch before custom ever ran. 
Fixed by building a dedicated router with only custom 
validation configured — no schema key. Also added 
transformRequest mention to the PR description. 
Pushed the fix.

PR #180, opened 2026-05-28
fix(crud): return 404 on DELETE of non-existent resource and fix validateBody crash

### LINKED ISSUE #130: validateBody crashes if not provided
# Bug Report 🐛

## Expected Behavior
If `validateBody` is not passed when creating a CRUD router, the POST handler 
should skip validation and insert the request body as-is.

## Current Behavior
The POST handler crashes with a TypeError at runtime because it accesses 
`validateBody.schema` without checking if `validateBody` itself exists.

## Possible Solution
Use optional chaining: change `if (validateBody.schema)` to `if (validateBody?.schema)`

## Steps to Reproduce
1. Create a router without passing `validateBody`:
   `buildCrudRouter({ table: myTable, typeName: 'contracts' })`
2. Mount it: `app.use('/contracts', router)`
3. Send a POST request to `/contracts` with a valid JSON body
4. Observe: `TypeError: Cannot read properties of undefined (reading 'schema')`

## Context (Environment)
This affects any route that uses `buildCrudRouter` without a validation schema,
which is a common case for simple tables that don't need input validation.

### Desktop
- OS: Any
- Browser: N/A (server-side)
- Version: Current main branch

## Detailed Description
In `buildCrudRouter`, the `validateBody` option is not marked as required in 
`CrudRouterOptions`, meaning callers are not expected to always pass it. However 
the POST handler accesses `validateBody.schema` unconditionally, causing a crash 
before any insert logic runs. This is a silent contract violation — the type 
allows omitting it but the runtime assumes it exists.

## Possible Implementation
// Before
if (validateBody.schema) {

// After
if (validateBody?.schema) {

Also consider marking it optional explicitly in the interface:
validateBody?: InsertValidator;

### LINKED ISSUE #131: DELETE /:id returns 200 even when the resource does not exist
# Bug Report 🐛
DELETE /:id returns 200 even when the resource does not exist

## Expected Behavior
When sending a DELETE request for an ID that does not exist in the database,
the server should return 404 { "error": "Not found" }.

## Current Behavior
The DELETE handler always returns 200 { "status": "deleted" } regardless of
whether the given ID matched any row. No error is thrown, no indication is
given that the resource was missing.

## Possible Solution
Use Drizzle's .returning() on the delete query to get back the deleted rows.
If the returned array is empty, no row was matched and the handler should
respond with 404 instead of 200.

## Steps to Reproduce
1. Set up a table using buildCrudRouter and mount it e.g. app.use('/contracts', router)
2. Pick an ID that does not exist in the table e.g. 99999
3. Send DELETE /contracts/99999
4. Observe: response is 200 { "status": "deleted" } — expected 404

## Context (Environment)
This affects every route built with buildCrudRouter since the DELETE handler
is generic. Any client relying on HTTP status codes to confirm deletion —
such as a UI showing a success toast or an automated cleanup script — will
behave incorrectly because it cannot distinguish a real deletion from a no-op.

### Desktop
- OS: Any
- Browser: N/A (server-side)
- Version: Current main branch

## Detailed Description
The current DELETE handler runs the delete query but never checks how many
rows were affected. Drizzle ORM's .returning() returns the deleted rows as
an array — if the array is empty, nothing was deleted and the handler should
return 404. This is already how PUT /:id behaves in the same codebase using
.returning(), so this fix makes DELETE consistent with the rest of the router.

### BEFORE
await res.locals.db
    .delete(table)
    .where(and(...whereConditions));

res.json({ status: 'deleted' });

### AFTER
const result = await res.locals.db
    .delete(table)
    .where(and(...whereConditions))
    .returning();

if (!result.length) {
    return res.status(404).json({ error: 'Not found' });
}

res.json({ status: 'deleted' });

Satvik77777:
## Summary
Fixes two bugs in the CRUD router and adds test coverage.
Supersedes closed PR #132 by @Ananya-vastare.

### Fix 1 — DELETE /:id returns 404 for non-existent resources
- Previously always returned 200 regardless of whether 
  the resource existed
- Now uses .returning() to check if anything was deleted
- Returns 404 with clear message if nothing was found

### Fix 2 — validateBody no longer crashes when not provided
- validateBody.schema was accessed without null check
- Changed to validateBody?.schema optional chaining

### Tests (9 passing)
- PUT schema validation failure returns 400
- PUT and POST apply identical validation
- Valid PUT body succeeds
- PUT custom Concerto validation failure returns 400
- Empty PUT body returns 400
- Custom validation runs without schema
- DELETE returns 404 when resource does not exist ← new
- DELETE returns 200 when resource exists ← new
- POST without validateBody does not crash ← new

## References
- Closes #130
- Closes #131
- Supersedes closed PR #132 by @Ananya-vastare

## Author Checklist
- [x] DCO sign-off provided
- [x] Bug fixes included
- [x] Tests added and passing (9/9)

### DIFF
--- a/server/handlers/crud.test.ts
+++ b/server/handlers/crud.test.ts
@@ -1,0 +1,114 @@
+import request from 'supertest';
+import express from 'express';
+import { jest } from '@jest/globals';
+import templatesRouter from './templates';
+import * as db from './templatebuilder';
+import * as validationModule from './concertovalidation';
+import { TemplateInsertSchema } from '../db/schema';
+
+jest.mock('../db/schema');
+jest.mock('./templatebuilder');
+jest.mock('./concertovalidation', () => {
+    return {
+        concertoValidation: jest.fn()
+    };
+});
+
+const mockedSchema = TemplateInsertSchema as jest.Mocked<typeof TemplateInsertSchema>;
+
+const validTemplateBody = {
+    uri: 'https://templates.accordproject.org/latedeliveryandpenalty@0.1.0.cta',
+    author: 'Accord Project',
+    displayName: 'Late Delivery And Penalty',
+    version: '0.1.0',
+    description: 'A template for late delivery',
+    license: 'Apache-2.0',
+    keywords: [] as string[],
+    logo: ''
+};
+
+describe('PUT /:id validation', () => {
+    let app: express.Application;
+
+    beforeEach(() => {
+        jest.clearAllMocks();
+        app = express();
+        app.use(express.json());
+        app.use((req, res, next) => {
+            res.locals.db = {
+                insert: jest.fn().mockReturnThis(),
+                values: jest.fn().mockReturnThis(),
+                returning: jest.fn<any>().mockResolvedValue([{ id: 1 }]),
+                update: jest.fn().mockReturnThis(),
+                set: jest.fn().mockReturnThis(),
+                where: jest.fn().mockReturnThis(),
+            };
+            next();
+        });
+        app.use('/templates', templatesRouter);
+    });
+
+    it('rejects invalid PUT body with 400 when schema validation fails', async () => {
+        (mockedSchema.safeParse as any) = jest.fn().mockReturnValue({ 
+            success: false, 
+            error: { errors: [{ message: 'Invalid body' }] } 
+        });
+        const res = await request(app)
+            .put('/templates/1')
+            .send({ invalid: 'data' });
+        expect(res.status).toBe(400);
+    });
+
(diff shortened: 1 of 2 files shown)


Satvik77777:
Hi @niallroche @mttrbrts — this PR fixes #130 and #131, 
both of which were left unresolved after PR #132 was 
auto-closed. Adds full test coverage (9/9 passing) for 
both fixes. DCO signed.

mttrbrts (approved):

mttrbrts:
The Windows build failure looks unrelated, so will merge and we can look at the build failure separately

PR #182, opened 2026-05-28
verifiable agreement signatures

sanketshevkar:


niallroche:
his is a good direction as discussed on a previous working group call @sanketshevkar  — the JCS canonicalization and the SIGNNG → SIGNING fix are both solid.

A couple of shape suggestions while it's still in draft, aimed at keeping the door open for more signature types later:

I would suggesting typing the hashes with the new org.accordproject.crypto@1.0.0 model that I proposed. It landed recently with ContentHash / HashAlgorithm / Canonicalization primitives (enum-plus-CUSTOM). Typing agreementDataHash / templateHash as crypto.ContentHash rather than bare strings makes the algorithm and the canonicalization scheme explicit and self-describing — which is exactly the property you want on a hash that a verifier re-derives. It also means the canonicalization choice is declared in the data instead of implied by the code.

Add a method discriminator to the signature record. The record is already nicely method-agnostic — verifiableCredential is optional and signatureImage is there for non-VC signatures — so tagging each record with the method that produced it (vc-did, sign-on-glass, …) costs almost nothing now and makes any future signature type an additive change rather than a migration to retrofit a discriminator onto existing rows.

Minor: [server/vendor/vc-signer-0.1.2.tgz](https://github.com/The-Building-Blocks/app-enterprise/blob/claude/merge-strategy-jayds22-f7gIZ/server/vendor/vc-signer-0.1.2.tgz) — vendoring the signing lib as an in-repo tarball is a supply-chain/maintenance consideration. If it can be a published dependency that'd be easier to audit and update; if it has to stay vendored for now, a one-line note on why would help future maintainers.

None of these are blocking — they're about keeping the record shape open as more methods and the crypto/agentic-commerce model family come online.

niallroche:
see https://github.com/accordproject/models/pull/195 @sanketshevkar

sanketshevkar:
Thanks @niallroche for the review, I had raised this as a draft PR for reviewing for the discussion we had started a couple of months back.

The tarball package included was a part of the demo, I still don't have a concrete answer on owning a utility specific for signing and validating VCs and DIDs specific to AP templates. Or maybe we'd just want to have it as a reference implementation in APAP.

JayDS22:
Not blocking, small heads-up from a scan today per @niallroche's Aug 5 endpoint-check ask:

The diff touches `model/protocol.cto` and adds `server/handlers/signatures.ts` + `server/handlers/vcSigner.ts`, but I did not see `openapi.json` or `openapi-extra-paths.json` in the changed-files list. If `signatures.ts` declares its own REST routes (or attaches new ones to the agreements router), the OpenAPI + generated TypeScript client will drift silently unless `npm run gen && npm run merge` runs at the repo root after the code changes settle.

Sequence I would follow once the current CONFLICTING state is rebased:
1. `cd apap && npm run gen` (Concerto `protocol.cto` -> `openapi.json`)
2. `cd apap && npm run merge` (folds `openapi-extra-paths.json` into `openapi.json` for non-CRUD verbs like sign / verify)
3. commit the regenerated `openapi.json` + `client/typescript/apap.ts` in the same PR

Happy to help wire the extra-paths entries if the signing surface exposes new REST verbs beyond what CRUD generation covers. Flagging early so it does not surface as a post-merge follow-up.

niallroche:
Following up now that the RI signing/freeze work has moved — a few coordination points so this doesn't collide with what landed and lands cleanly on top of it.

**1. Drop the `SIGNNG → SIGNING` typo fix and rebase on #240.** #240 fixes the same enum typo the complete way — model source → generated artifacts → a migration with `ALTER TYPE … RENAME VALUE` to preserve existing data. Keeping the fix here will conflict; once #240 lands, rebase and drop it.

**2. #240's record-freeze is the substrate this needs — build on it, not parallel to it.** #240 freezes `data` at SIGNING and full-freezes at COMPLETED, which is exactly what makes a signature meaningful (a signature over still-mutable content isn't verifiable). So sequence this after #240 and let the frozen record be what the attestation binds to.

**3. Align on one canonical content hash.** #238 introduces `atrHash` (byte-pinned SHA-256 over the frozen agreement); there's also the `crypto@1.0.0.ContentHash` binding I suggested here. These should be one definition — same algorithm and canonicalization — so the signature, `atrHash`, and the LCP terms hash all bind the same bytes. Three different "the agreement is this" digests would be a latent interop bug.

Still open from the earlier review, and worth doing as part of the rebase:
- Type the content hash via `org.accordproject.crypto@1.0.0.ContentHash` rather than a bare string (this is the same point as #3).
- Add a `SignatureMethod` discriminator so `vc-did` is one method among future ones, keeping new methods additive.
- The vendored `vc-signer-0.1.2.tgz` is a supply-chain concern — publish it as a dependency or document why in-repo vendoring is necessary.
- @JayDS22's note stands too: regen `openapi.json` / `openapi-extra-paths.json` after the rebase so the TS client stays in sync.

Happy to review in depth once it's rebased on #240 with the hash definition settled — that's the piece that unblocks the rest.

niallroche:
keeping alive

PR #184, opened 2026-06-04
feat(server): introduce typed ServiceError hierarchy, wire agreements handler

JayDS22:
First slice of the GSoC W1 service-layer track from the project roadmap. Ports the typed-error hierarchy from the [apap-mcp-poc](https://github.com/JayDS22/apap-mcp-poc) into `apap/server`. Marked draft because remaining handlers (templates, mcp, the inner crud catch) will be migrated in subsequent PRs, but this one is reviewable and tested as-is.

## Why

Handlers in this repo throw bare `Error` strings everywhere:

```ts
throw new Error(`Agreement with id ${agreementId} does not exist`);
```

That gives route catch blocks no way to distinguish a 404 from a 500, and gives clients nothing actionable in the response body. Every failure ends up as a generic 500 with a string message, even when the right answer is a 404 with structured details.

## What

**`server/services/errors.ts`** (new, 105 lines): `ServiceError` base class + 6 typed subclasses, each carrying a machine-readable code, an HTTP status, a human message, and optional structured details:

| Class | Code | Status |
|---|---|---|
| `TemplateNotFoundError` | `TEMPLATE_NOT_FOUND` | 404 |
| `TemplateDuplicateError` | `TEMPLATE_DUPLICATE` | 409 |
| `AgreementNotFoundError` | `AGREEMENT_NOT_FOUND` | 404 |
| `AgreementConversionError` | `AGREEMENT_CONVERSION_FAILED` | 500 |
| `InvalidPayloadError` | `INVALID_PAYLOAD` | 400 |
| `ValidationError` | `VALIDATION_ERROR` | 422 |

`toJSON()` emits the wire shape. `Object.setPrototypeOf(this, new.target.prototype)` in the constructor keeps `instanceof` working after transpilation.

**`server/handlers/agreements.ts`** (wiring as proof point):
- `resolveAgreement` now throws `AgreementNotFoundError` when the agreement row is missing
- `/:id/convert/:format` and `/:id/trigger` outer catch blocks recognise `ServiceError` and map it to `error.statusCode + error.toJSON()`. Bare `Error` throws still fall through to the existing 500 path, so this is additive at the handler boundary

## Demo

Before:

\`\`\`
\$ curl -i http://localhost:9000/agreements/9999/convert/html
HTTP/1.1 500 Internal Server Error
{\"error\":\"Agreement with id 9999 does not exist\"}
\`\`\`

After:

\`\`\`
\$ curl -i http://localhost:9000/agreements/9999/convert/html
HTTP/1.1 404 Not Found
{
  \"error\": {
    \"code\": \"AGREEMENT_NOT_FOUND\",
    \"message\": \"Agreement not found: 9999\",
    \"details\": { \"identifier\": \"9999\" }
  }
}
\`\`\`

## Tests

- `server/services/errors.test.ts` (new, 12 tests): contract tests for the 6 typed errors plus the `ServiceError` base. Covers `toJSON` shape with/without details, statusCode mapping, details propagation, and instanceof chain across all subclasses.
- `server/handlers/agreements.test.ts` (1 updated): the existing assertion that "agreement not found" returns 500 with a bare string is updated to assert the new 404 + structured body. That assertion was itself a bug, the response was always supposed to be a 404.

Full suite: **59/59 passing**, tsc clean.

## Scope intentionally limited

Only agreements.ts is wired in this PR. Templates, mcp, and the inner crud catch will follow in subsequent PRs to keep this one reviewable. The intent is that once `ServiceError` lives in `server/services/`, every later refactor PR can throw/catch it without re-litigating the wire shape.

## Related

Part of the GSoC W1 deliverable per the project roadmap. Aligns with [apap-mcp-poc](https://github.com/JayDS22/apap-mcp-poc) which has been running this pattern since the bonding period (53 tests, 98.5% coverage). Service-layer modules (`templateService.ts`, `agreementService.ts`) will follow in their own PRs through W1-W2.

### DIFF
--- a/server/services/errors.test.ts
+++ b/server/services/errors.test.ts
@@ -1,0 +1,101 @@
+import {
+    ServiceError,
+    TemplateNotFoundError,
+    TemplateDuplicateError,
+    AgreementNotFoundError,
+    AgreementConversionError,
+    InvalidPayloadError,
+    ValidationError,
+} from './errors';
+
+describe('ServiceError', () => {
+    it('exposes code, statusCode, message, and details', () => {
+        const err = new ServiceError('CUSTOM_CODE', 418, 'tea time', { teapot: true });
+        expect(err.code).toBe('CUSTOM_CODE');
+        expect(err.statusCode).toBe(418);
+        expect(err.message).toBe('tea time');
+        expect(err.details).toEqual({ teapot: true });
+    });
+
+    it('omits details from toJSON when not provided', () => {
+        const err = new ServiceError('NO_DETAILS', 500, 'boom');
+        expect(err.toJSON()).toEqual({
+            error: { code: 'NO_DETAILS', message: 'boom' },
+        });
+    });
+
+    it('includes details in toJSON when provided', () => {
+        const err = new ServiceError('WITH_DETAILS', 400, 'bad', { field: 'name' });
+        expect(err.toJSON()).toEqual({
+            error: {
+                code: 'WITH_DETAILS',
+                message: 'bad',
+                details: { field: 'name' },
+            },
+        });
+    });
+
+    it('survives prototype chain for instanceof after transpilation', () => {
+        const err = new TemplateNotFoundError('abc');
+        expect(err).toBeInstanceOf(TemplateNotFoundError);
+        expect(err).toBeInstanceOf(ServiceError);
+        expect(err).toBeInstanceOf(Error);
+    });
+});
+
+describe('Template errors', () => {
+    it('TemplateNotFoundError -> 404 TEMPLATE_NOT_FOUND with identifier in details', () => {
+        const err = new TemplateNotFoundError(42);
+        expect(err.statusCode).toBe(404);
+        expect(err.code).toBe('TEMPLATE_NOT_FOUND');
+        expect(err.details).toEqual({ identifier: 42 });
+        expect(err.message).toContain('42');
+    });
+
+    it('TemplateDuplicateError -> 409 TEMPLATE_DUPLICATE with uri in details', () => {
+        const err = new TemplateDuplicateError('resource:foo#bar');
+        expect(err.statusCode).toBe(409);
+        expect(err.code).toBe('TEMPLATE_DUPLICATE');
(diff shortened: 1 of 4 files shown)


server/services/errors.test.ts
```diff
@@ -0,0 +1,101 @@
...
+            },
+        });
+    });
+
+    it('survives prototype chain for instanceof after transpilation', () => {
+        const err = new TemplateNotFoundError('abc');
+        expect(err).toBeInstanceOf(TemplateNotFoundError);
+        expect(err).toBeInstanceOf(ServiceError);
+        expect(err).toBeInstanceOf(Error);
+    });
```
mttrbrts, line 44: Nice!

server/services/errors.test.ts
```diff
@@ -0,0 +1,101 @@
...
+        expect(err.statusCode).toBe(400);
+        expect(err.code).toBe('INVALID_PAYLOAD');
+        expect(err.details).toEqual({ field: 'uri' });
+    });
+
+    it('ValidationError -> 422 VALIDATION_ERROR', () => {
+        const err = new ValidationError('schema mismatch');
+        expect(err.statusCode).toBe(422);
+        expect(err.code).toBe('VALIDATION_ERROR');
+    });
```
mttrbrts, line 102: Nit: Also test that err.details is undefined/empty?

mttrbrts (approved):
A minor observation, on the tests. Tag me when you're ready and I'll merge. The other handlers can come later.

JayDS22:
@mttrbrts addressed the details-undefined nit in `9e794f6`. Added `expect(err.details).toBeUndefined()` to both the `omits details from toJSON when not provided` test (existing test, now asserts the property too) and the `ValidationError -> 422` test (where you flagged it).

Full suite still 59/59 green locally. Ready to merge whenever you have a moment. Thanks for the quick turnaround on the review.

JayDS22:
Thanks @mttrbrts for the quick turnaround. Will tackle the remaining handlers (templates, mcp, crud inner catch) as a follow-up PR bundling two per Niall's earlier preference. Targeting this week.

PR #187, opened 2026-06-04
fix(crud): prevent SQL injection via identifier validation and operator whitelist

Satvik77777:
## Summary
Fixes SQL injection vulnerability in defaultWhereClause 
where query parameter keys and operators were passed 
directly into sql.raw() without sanitization.

### Fix 1 — Prototype pollution prevention
Replaces `key in table` with 
`Object.prototype.hasOwnProperty.call(table, key)` 
to block prototype keys like toString, __proto__, 
constructor from reaching sql.raw()

### Fix 2 — Column identifier validation
Validates column names against SAFE_IDENTIFIER_RX 
(/^\w+$/) before passing to sql.raw() — rejects any 
key containing special characters

### Fix 3 — Operator whitelist
Validates SQL operators against explicit ALLOWED_OPERATORS 
array instead of passing regex-matched user input 
directly to sql.raw()

### Fix 4 — Cleanup
Removes unused escapeString import (designed for SQL 
string values, not identifiers)

### Tests (6 new, 15 total passing)
- toString prototype pollution key → rejected
- __proto__ key → rejected
- constructor key → rejected
- Unknown column key → rejected
- Operator filter syntax → works correctly
- Null filter value → handled safely

## Security Impact
Without this fix an attacker controlling query parameter 
key names could inject arbitrary SQL into column 
references via sql.raw()

## References
- Closes #186

## Author Checklist
- [x] DCO sign-off provided
- [x] Security fix
- [x] Tests added and passing (15/15)

PR #192, opened 2026-06-06
refactor(error-handling): centralize duplicate try/catch into Express error middleware

### LINKED ISSUE #191: Title: refactor(error-handling): centralize duplicate  try/catch blocks into Express error middleware
The same try/catch -> console.log -> res.status(500) 
pattern is duplicated 20+ times across agreements.ts, 
crud.ts, and mcp.ts. Several console.log statements 
also expose PII (req.body, contract data) and internal 
details (SQL queries, stack traces) to stdout.

This issue tracks:
1. Creating server/middleware/errorHandler.ts with 
   globalErrorHandler and asyncHandler utilities
2. Refactoring all route handlers to use asyncHandler
3. Removing PII-exposing console.log statements
4. Replacing with structured non-leaking error logging

Satvik77777:
## Summary
The same try/catch → console.log → res.status(500) pattern 
was duplicated 20+ times across agreements.ts, crud.ts, 
and mcp.ts. Several console.log statements also exposed 
PII and internal details to stdout. This PR centralizes 
error handling into a single Express error middleware.

### New file — server/middleware/errorHandler.ts
- globalErrorHandler — single centralized error response 
  location, replaces 20+ duplicate catch blocks
- asyncHandler — wraps async route handlers to forward 
  errors via next() without try/catch in every route
- toSafeMessage — prevents stack traces and internal 
  paths from leaking to API clients

### Refactor — agreements.ts
- Wrapped async route handlers with asyncHandler
- Removed duplicate try/catch blocks
- Removed console.log statements exposing:
  * req.body (user PII)
  * agreement.data and agreement.state (contract data)
  * err.stack (internal stack traces)

### Refactor — crud.ts
- Wrapped async route handlers with asyncHandler
- Removed duplicate try/catch blocks
- Removed console.log(sql.sql) and console.log(sql.params)
  which leaked raw SQL query structure to stdout

### Refactor — mcp.ts
- Removed console.log(req.body) on every MCP POST request
- Replaced with structured console.error using 
  non-sensitive context only

### Security
Raw contract data, SQL queries, and stack traces were 
being logged to stdout. In any hosted environment these 
appear in server logs accessible to operators and 
potentially leaked in log aggregation tools.

### Maintainability
20+ identical catch blocks meant any change to error 
handling required editing every handler file. Now 
there is one place to update.

### Reliability
asyncHandler ensures unhandled promise rejections in 
route handlers are always caught and forwarded to 
Express error handling — preventing silent process 
crashes in Node.js.

## References
- Closes #191
- Coordinates with #184 — middleware honours ServiceError statusCode and toJSON() shape introduced there.

## Author Checklist
- [x] DCO sign-off provided
- [x] New middleware file with JSDoc
- [x] All existing tests passing
- [x] TypeScript compilation verified (npx tsc --noEmit)
- [x] ServiceError instances return typed status codes and structured body via #184 integration

### DIFF
--- a/server/handlers/agreements.test.ts
+++ b/server/handlers/agreements.test.ts
@@ -10,6 +10,7 @@ import AdmZip from 'adm-zip';
 import fs from 'fs';
 import path from 'path';
 import * as validationModule from './concertovalidation';
+import { globalErrorHandler } from '../middleware/errorHandler';
 jest.setTimeout(30000);
 
 // Mock dependencies (but not TemplateArchiveProcessor)
@@ -97,6 +98,7 @@ describe('Agreements Router - POST /:id/trigger', () => {
 
         // Add agreements router
         app.use('/agreements', agreementsRouter);
+        app.use(globalErrorHandler);
     });
 
     describe('Success scenarios', () => {
@@ -472,8 +474,9 @@ describe('Agreements Router - POST /:id/trigger', () => {
                 .send('{ invalid json }')
                 .expect(400);
 
-            // Express default error handling for malformed JSON
-            expect(response.text).toContain('SyntaxError');
+            // globalErrorHandler formats the JSON parsing error
+            expect(response.text).toContain('error');
+            expect(response.body.error).toBeDefined();
         });
     });
 
--- a/server/handlers/agreements.ts
+++ b/server/handlers/agreements.ts
@@ -7,7 +7,8 @@ import { eq } from 'drizzle-orm';
 import { TemplateArchiveProcessor } from '@accordproject/template-engine';
 import { HttpTemplateRetriever } from './retrievers/HttpTemplateRetriever';
 import { Template as CiceroTemplate } from '@accordproject/cicero-core';
-import { ServiceError, AgreementNotFoundError } from '../services/errors';
+import { AgreementNotFoundError } from '../services/errors';
+import { asyncHandler } from '../middleware/errorHandler';
 
 /**
  * @param db The database handle stored on `res.locals.db`.
@@ -64,8 +65,7 @@ const router = express.Router();
  * resolves and caches a remote template archive when the template URI matches a supported
  * retriever, and finally inserts the agreement into the database.
  */
-router.post('/', async (req, res) => {
-    try {
+router.post('/', asyncHandler(async (req, res) => {
         const db = res.locals.db;
         
         const zodValidation = AgreementInsertSchema.safeParse(req.body);
@@ -119,11 +119,7 @@ router.post('/', async (req, res) => {
 
         const inserted = await db.insert(Agreement).values(insertData).returning();
         res.json(inserted[0]);
-
-    } catch (err: any) {
-        res.status(500).json({ error: "Agreement Instantiation Failed", details: err.message });
-    }
-});
+}));
 
 const crudRouter = buildCrudRouter({
     table: Agreement,
@@ -138,24 +134,12 @@ const crudRouter = buildCrudRouter({
  * @details Resolves the agreement and its template, creates a `TemplateArchiveProcessor`,
  * and delegates the conversion to the template engine's draft support for the requested format.
  */
-crudRouter.get('/:id/convert/:format', async function (req, res) {
-    try {
+crudRouter.get('/:id/convert/:format', asyncHandler(async function (req, res) {
         const {agreement, apTemplate} = await resolveAgreement(res.locals.db, req.params.id);
         const templateArchiveProcessor = new TemplateArchiveProcessor(apTemplate);
         const draftResult = await templateArchiveProcessor.draft(agreement.data, req.params.format, {});
-        console.log(draftResult);
-        res.setHeader("Content-Type", `text/${req.params.format}`);
         res.send(draftResult);
-    } catch (error) {
-        console.log(error);
-        if (error instanceof ServiceError) {
-            res.status(error.statusCode).json(error.toJSON());
-            return;
-        }
-        const message = error instanceof Error ? error.message : 'Unknown error';
-        res.status(500).json({ error: message });
-    }
-});
+}));
 
 /**
  * @param req The Express request containing the agreement id and trigger payload.
@@ -165,14 +149,10 @@ crudRouter.get('/:id/convert/:format', async function (req, res) {
  * initializes agreement state when needed, executes the agreement logic through the
  * template archive processor, and persists the updated agreement state back to the database.
  */
-crudRouter.post('/:id/trigger', async function (req, res) {
-    try {
+crudRouter.post('/:id/trigger', asyncHandler(async function (req, res) {
         const {agreement, apTemplate} = await resolveAgreement(res.locals.db, req.params.id);
         const templateArchiveProcessor = new TemplateArchiveProcessor(apTemplate);
         try {
-            console.log(JSON.stringify(req.body));
-            console.log(JSON.stringify(agreement.data));
-            console.log(JSON.stringify(agreement.state));
 
             const requestSchema = apTemplate.getRequestTypes().find((rt: any) => rt === req.body.$class);
             if (!requestSchema) {
@@ -194,28 +174,20 @@ crudRouter.post('/:id/trigger', async function (req, res) {
             
             const triggerResult = await templateArchiveProcessor.trigger(agreement.data, req.body, agreement.state);
             agreement.state = triggerResult.state;
-            console.log(JSON.stringify(triggerResult));
 
             // Persist updated state.
             await res.locals.db.update(Agreement).set(agreement).where(eq(Agreement.id, Number.parseInt(agreement.id)));
             res.json(triggerResult);
         } catch (err: any) {
-            console.log("\n=== TRIGGER EXECUTION ERROR ===");
-            console.log(err.stack);
-            console.log("===============================\n");
+            console.error({ 
+                type: 'operation_failed', 
+                operation: 'triggerAgreement',
+                agreementId: req.params.id 
+            });
             
             res.json({ isError: true, errorMessage: err.message, errorDetails: err.toString() });
         }
-    } catch (error) {
-        console.log(error);
-        if (error instanceof ServiceError) {
-            res.status(error.statusCode).json(error.toJSON());
-            return;
-        }
-        const message = error instanceof Error ? error.message : 'Unknown error';
-        res.status(500).json({ error: message });
-    }
-});
+}));
 
 router.use('/', crudRouter);
 export default router;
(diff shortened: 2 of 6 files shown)


Satvik77777:
Hi @niallroche @mttrbrts — this PR centralizes the 
duplicate try/catch pattern (repeated 20+ times across 
agreements.ts, crud.ts, mcp.ts) into a single Express 
error middleware. Also removes PII-exposing console.log 
statements that were logging raw contract data and SQL 
queries to stdout. All tests passing, DCO signed.

JayDS22:
Hi @Satvik77777, nice PR. The middleware pattern + asyncHandler wrapper are a real improvement over the duplicate try/catch noise.

Flagging an overlap with my #184 that I'd like to coordinate on: that PR introduces a typed `ServiceError` hierarchy (`TemplateNotFoundError` 404, `AgreementNotFoundError` 404, `ValidationError` 422, etc.) so handlers can throw typed errors and have route catches map them to the right HTTP status + structured body. As written, your middleware will catch every error and respond 500, which would erase the per-error-class status mapping #184 sets up.

The two approaches compose nicely though: keep your middleware as the plumbing (`asyncHandler` + central error response location), and have the middleware check `instanceof ServiceError` to honour `error.statusCode` and `error.toJSON()`. Routes lose their try/catch (your improvement), but the typed errors keep their semantic richness (#184's contribution).

Would you be open to either:

(a) waiting for #184 to land first, then rebasing #192 on top of it with the `instanceof` check added to the middleware, or

(b) folding the `ServiceError` check into #192 directly so they ship together?

Happy to pair on the rebase if useful. Tagging @niallroche for direction.

Satvik77777:
Hi @JayDS22 — thanks for the detailed flag, I'd prefer option (b) — folding the instanceof ServiceError check directly into globalErrorHandler in this PR. I'll pull errors.ts from #184, wire the check into the middleware, and update the PR description to reference the coordination. That way #192 ships as a complete solution that already has typed error status codes, and your subsequent PRs just need to throw the right typed errors — no middleware changes needed.
Tagging @niallroche for direction as you suggested.

JayDS22:
Thanks @Satvik77777, option (b) sounds good. Quick status: #184 is approved and queued to merge today (just pushed the requested test nit on errors.test.ts), so the rebase target should be there shortly.

Once #184 lands, the `ServiceError` classes will live at `server/services/errors.ts`. The middleware addition would look something like:

```ts
import { ServiceError } from '../services/errors';

export function globalErrorHandler(err, req, res, next) {
  if (err instanceof ServiceError) {
    return res.status(err.statusCode).json(err.toJSON());
  }
  // fall through to existing generic 500 handling
  ...
}
```

That preserves the per-error-class status mapping (404 / 409 / 422 / etc) while keeping the asyncHandler + centralized response location your PR introduces. Happy to pair on the rebase if useful, or review when you push.

Satvik77777:
@JayDS22 — rebased on main after #184 merged, instanceof ServiceError check wired into globalErrorHandler. Also addressed Copilot nits: import moved to top of index.ts, console.error → console.log for routine MCP log, removed unused sql variable in crud.ts. 64/64 tests passing, tsc clean.

JayDS22:
@Satvik77777 #184 is in main. The `ServiceError` classes now live at `server/services/errors.ts`. Rebase target is ready whenever you want to fold the `instanceof` check into the middleware.

Satvik77777:
@JayDS22 — already done, see my comment above. Rebased on main, instanceof ServiceError wired into globalErrorHandler, 64/64 tests passing.

JayDS22:
@Satvik77777 nice turnaround on the rebase. The middleware + asyncHandler composition looks right. The `instanceof ServiceError` check preserves the typed status codes from #184, and the catch-block consolidation is a clear improvement.

Heads-up on merge order: my slice 2 (#197, draft) also touches `agreements.ts` (the three throws in `resolveAgreement`) and `templatebuilder.ts`. The throws compose cleanly with your middleware since typed errors bubble up automatically, but the file diff will conflict at whichever PR merges second. Happy to rebase #197 on top of #192 when yours lands; it's smaller and easier to rebase.

JayDS22:
@Satvik77777 picking up @niallroche's notes on this PR from the merge plan thread, two regressions worth fixing before merge:

**Content-Type lost on `/:id/convert/:format`** (`server/handlers/agreements.ts`)
The handler previously set `res.setHeader("Content-Type", \`text/${req.params.format}\`)` before the send. That line is gone after the asyncHandler wrap and there is no replacement, so the conversion endpoint will respond with the default JSON Content-Type and break HTML / markdown / pdf clients. Fix is to re-add the setHeader call inline before the send.

**23505 detail dropped the typeName context** (`server/handlers/crud.ts`)
The pre-refactor code returned `details: \`A resource with this unique identifier already exists for ${typeName}.\`` on a unique-constraint violation. The new `globalErrorHandler` handles 23505 generically but has no `typeName` to interpolate, so the API response now drops the resource-type detail. Either attach `typeName` to the error before forwarding (`error.typeName = typeName`) and read it in the middleware, or keep the detail-shaping in the route and only forward unexpected errors.

Otherwise the asyncHandler + globalErrorHandler pattern itself is the right idiom, just these two gaps to close.

Satvik77777:
Fixed both regressions:
- Content-Type header restored on /:id/convert/:format before send
- typeName now attached to 23505 errors in insert/update routes and interpolated in globalErrorHandler response

JayDS22:
Nice, thanks for wiring the middleware to honor #184's `ServiceError.toJSON()` and `.statusCode`. That was the coordination question I was about to ask before opening my next service-layer PR.

Small merge-order heads-up: my #208 (strict-numeric `:id` guard) adds a `router.param('id', ...)` block to `crud.ts` right after the `res.locals.orgId` middleware. Your PR reworks catch blocks in the same file. Different regions so a rebase should be clean either way, just flagging so whoever lands second knows where the diff lands.

niallroche:
hi @Satvik77777 — thanks for this, the structure is solid: the globalErrorHandler middleware, asyncHandler wrapping, and the agreements.ts / crud.ts cleanup all look good, and it composes fine with the typed ServiceError work from #200 and the session cleanup from #194.

One thing to fix before it lands, and it's a consequence of how long the branch has been open rather than anything wrong with your changes: the branch predates #200, which merged afterward and rewrote getAgreement in mcp.ts. In doing so it reintroduced PII logging that your diff can't see, so the merged result still leaks agreement contents to stdout — the exact thing this PR is meant to close. On current main's [server/handlers/mcp.ts](https://github.com/The-Building-Blocks/app-enterprise/blob/claude/merge-strategy-jayds22-f7gIZ/server/handlers/mcp.ts):

console.log(`Fetching agreement with ID: ${agreementId}`);
console.log(`Successfully fetched agreement: ${JSON.stringify(agreement)}`);
That second line dumps the full agreement JSON. There are ~28 console.* calls in mcp.ts on current main; your original removal only covered the ones that existed at your June base.

Could you rebase onto current main and extend the PII-log sweep to cover mcp.ts's post-[#200](https://github.com/The-Building-Blocks/app-enterprise/issues/200) logging as well? Once the agreement/template-content logs are gone from mcp.ts too, the security claim in the PR body holds end-to-end and this is good to go.

One process note that'll save us a repeat of the #194 near-miss: when you sync, please use the "Update branch" button on the PR (or a rebase) rather than a local git merge main with hand-resolved conflicts — that's how #187's guards briefly got dropped on [#194](https://github.com/The-Building-Blocks/app-enterprise/issues/194) before you caught it. CI can reach models.accordproject.org so it'll give the real green on the agreements suite once you push.

Satvik77777:
@niallroche — rebased onto current main (not a local merge, per your note) and audited all ~26 console.* calls in mcp.ts.
PII logging fixes:

Fixed the two lines you specifically flagged — getAgreement's fetch/success logs no longer dump the full agreement JSON
Also found and removed a raw req.body dump on every incoming MCP request (console.log(JSON.stringify(req.body, null, 2))) — wasn't in the original scope but leaks the same class of data
All console.error calls across getAgreement, getTemplates, getAgreements, draftAgreement, triggerAgreement, and the tool handlers now log structured, non-sensitive metadata (operation name, resource ID, HTTP status) instead of raw API response bodies. The full error is still thrown to the calling client — only the server-side log is stripped of sensitive content.
Removed the raw request-body log in triggerAgreement (was logging transaction payloads directly)

One correctness fix along the way: a cleanup-failure log (transport_close_error) had been accidentally downgraded from console.error to console.log during the refactor — restored it to console.error since it represents a genuine failure condition, not routine activity.
64/64 tests passing after rebase and after the logging changes.

niallroche (changes requested):
very good restructuring, there is a small comment copilot caught for index.ts line 77 around imports to address.

This introduces a static import mid-file after executable statements. While still top-level, many TS/ESLint configs enforce imports at the top for consistent module initialization and to avoid tooling edge-cases. Move the globalErrorHandler import to the existing import section at the top of the file.

Satvik77777:
@niallroche — checked this carefully. The globalErrorHandler import is at line 2, grouped with the other top-level imports. Line 77 is app.use(globalErrorHandler) — the middleware registration call, correctly placed after all routes (not an import statement). This appears to already be resolved on the current pushed commit. Let me know if you'd like me to push a no-op commit to refresh CI, or if a fresh look at the current diff would confirm it on your end.

JayDS22 (approved):
APPROVE. The middleware honors `ServiceError.statusCode` / `toJSON()` from #184, and the Jul 17 PII-log ask from @niallroche is addressed on the Jul 18 push (raw `req.body` dump gone from `/mcp`, `agreement.data`/`.state` dumps gone from the trigger inner catch, SQL query logging gone from crud.ts list). My #208 `router.param('id', ...)` guard and Satvik's `SAFE_IDENTIFIER_RX` (from #187) survived the rebase cleanly.

Two optional follow-ups I would defer, not block:

1. No direct tests for `middleware/errorHandler.ts`. ServiceError mapping is covered indirectly by existing agreement-handler tests and the malformed-JSON path is covered by the appended `agreements.test.ts` case, but the 23505 branch and the `err.status` extraction path are only exercised end-to-end. A small `errorHandler.test.ts` with three cases (23505 -> 409 with typeName, ServiceError -> statusCode/toJSON, generic error -> 500) would pin the contract.

2. Original postgres error message lost on 23505 re-throw in `crud.ts`. The pattern in POST + PUT creates `new Error('Unique constraint violation')` and attaches `code`/`typeName`, dropping the underlying detail like "duplicate key value violates unique constraint 'template_uri_key'". Client response is unchanged, but internal logs lose the constraint name. `new Error('Unique constraint violation', { cause: error })` (ES2022 error cause, works on Node 20+) would preserve it without changing the client contract.

niallroche (approved):
Approving and merging. Thanks @Satvik77777 for staying with this one — it's been open a while and landed clean.

Confirmed on the current diff:

globalErrorHandler correctly registered after routes (import at line 2, app.use at line 77) — the inline-import concern was a stale earlier view, current placement is right
PII sweep is thorough: raw req.body dump gone from /mcp, agreement.data/.state gone from the trigger catch, SQL query logging gone from crud.ts, and the mcp.ts logs are now structured metadata-only (id + status + count, no content)
Middleware honors ServiceError.statusCode/toJSON() from #184
[#187](https://github.com/The-Building-Blocks/app-enterprise/issues/187)'s SAFE_IDENTIFIER_RX and #208's router.param('id') guard both survived the rebase intact

Thanks @JayDS22 for the review

PR #189, opened 2026-06-05
test(mcp): add test suite for mcp.ts handler and fix session header bug

Satvik77777:
## Summary
Adds the first test suite for mcp.ts which previously had 0% test 
coverage. The session-header bug fix and session cleanup discovered 
during testing have been split into separate PRs (#203 and #194) per 
@niallroche's suggestion.

## What this PR does

**Refactor — export getServer() for testability**
- Exports getServer() as a named export so tools and resources can be 
  tested without fighting the StreamableHTTPServerTransport stream 
  mechanics

**Tests (7 new, 60 total passing)**
- POST /mcp without session ID returns 400
- POST /messages with bad session ID returns 400
- getTemplate tool is registered correctly
- getAgreement tool is registered correctly
- trigger-agreement tool is registered correctly
- Tools tested via InMemoryTransport from official MCP SDK — bypasses 
  HTTP stream mechanics entirely
- Full baseline (53 tests) still passing

## Why InMemoryTransport over supertest
StreamableHTTPServerTransport requires specific Accept headers and 
stream mechanics that supertest cannot emulate. Using InMemoryTransport 
from the official @modelcontextprotocol/sdk tests the actual tool logic 
with 100% confidence.

## References
- Closes #188
- Depends on #203 (session-header fix) and #194 (session cleanup) 
  landing first
- Supports RFC #143 (shared service layer refactor)
- Related to GSoC project on APAP/MCP hardening

## Author Checklist
- DCO sign-off provided
- Tests added and passing (60/60)
- No behavior changes

PR #194, opened 2026-06-07
fix(mcp): add TTL-based session cleanup to prevent memory leak

### LINKED ISSUE #193: fix(mcp): transports map grows indefinitely —  memory leak on unclean client disconnect
The global transports map in mcp.ts stores active MCP 
sessions but has no TTL or cleanup mechanism. Sessions 
are only removed when transport.onclose fires correctly.
If a client disconnects uncleanly (network drop, crash, 
timeout), the transport remains in the map forever.

Over time this causes unbounded memory growth that will 
crash the server in production.

Proposed fix:
- Add a lastActivity timestamp per session
- Run a periodic cleanup interval every 5 minutes
- Remove sessions idle for more than 30 minutes

Satvik77777:
## Summary
The global transports map in mcp.ts had no cleanup 
mechanism. Sessions were only removed when 
transport.onclose fired correctly. Unclean client 
disconnects (network drops, crashes, timeouts) left 
transports in the map forever, causing unbounded 
memory growth that crashes the server in production.

### New cleanup mechanism
- sessionLastActivity map tracks last request 
  timestamp per session
- setInterval runs every 5 minutes removing sessions 
  idle for 30+ minutes
- interval.unref() prevents blocking process exit

### Activity tracking
- lastActivity updated on every incoming request 
  for existing sessions
- lastActivity set on onsessioninitialized callback

### Cleanup on disconnect
- sessionLastActivity cleaned up on transport.onclose
- sessionLastActivity cleaned up on SSE res.on("close")
- structured console.error logging for cleanup events

### Tests (3 new, 14 total passing)
- tracks session activity on initialization
- cleanup interval registered with unref
- session removed from both maps on close

## References
- Closes #193

## Author Checklist
- [x] DCO sign-off provided
- [x] Memory leak fix
- [x] Tests added and passing (14/14)
- [x] TypeScript compilation verified

### DIFF
--- a/server/handlers/mcp-session-cleanup.test.ts
+++ b/server/handlers/mcp-session-cleanup.test.ts
@@ -1,26 +1,37 @@
-import { transports, sessionLastActivity, sessionCleanupInterval } from './mcp';
+import { transports, sessionLastActivity, sessionCleanupInterval, startSessionCleanup } from './mcp';
 
 describe('MCP Session Cleanup Test', () => {
+    beforeEach(() => {
+        jest.useFakeTimers();
+        startSessionCleanup();
+    });
+
+    afterEach(() => {
+        jest.useRealTimers();
+        jest.restoreAllMocks();
+        // Clear all keys from maps to avoid test pollution
+        for (const key of Object.keys(transports)) {
+            delete transports[key];
+        }
+        for (const key of Object.keys(sessionLastActivity)) {
+            delete sessionLastActivity[key];
+        }
+    });
+
     it('tracks session activity on initialization', () => {
         const testId = 'activity-test-session';
         sessionLastActivity[testId] = Date.now();
         expect(typeof sessionLastActivity[testId]).toBe('number');
         expect(sessionLastActivity[testId]).toBeLessThanOrEqual(Date.now());
-        delete sessionLastActivity[testId]; // cleanup
     });
 
     it('cleanup interval is registered with unref', () => {
-        // sessionCleanupInterval should exist and 
-        // not block process exit
         expect(sessionCleanupInterval).toBeDefined();
         expect(typeof (sessionCleanupInterval as any).hasRef).toBe('function');
         expect((sessionCleanupInterval as any).hasRef()).toBe(false);
     });
 
     it('session is removed from both maps on close', () => {
-        // When transport.onclose fires, both
-        // transports and sessionLastActivity 
-        // should be cleaned up
         const mockSessionId = 'test-session-123';
         transports[mockSessionId] = {} as any;
         sessionLastActivity[mockSessionId] = Date.now();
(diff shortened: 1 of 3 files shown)


Satvik77777:
@niallroche @mttrbrts — this PR fixes a memory 
leak where unclean MCP client disconnects left 
sessions in the transports map forever. Added 
TTL-based cleanup running every 5 minutes with 
30 minute idle timeout. 14/14 tests passing, 
DCO signed. Directly relevant to @JayDS22 's GSoC 
work on MCP hardening

Satvik77777:
Hi @niallroche — flagging before this goes stale/closes: this branch 
was cut before the #189 split, so it still carries the test-suite and 
header-fix commits (11339a6, 65930b9) alongside the actual cleanup 
fix (a13b34e). Those first two are now duplicated in #203 and #189.

I'll rebase this branch to keep only the cleanup commit (a13b34e), 
so #194 is scoped to just the memory-leak fix as you described.

Satvik77777:
Rebased to isolate just the cleanup fix. Dropped the test-suite/
header-fix commits (11339a6, 65930b9) that were redundant with #203/
#189. New minimal test file (mcp-session-cleanup.test.ts) with 3 
passing tests covering the memory-leak fix only. Ready for review.

JayDS22 (commented):
Nice catch on the leak. Session cleanup was overdue. A few things worth thinking about before this lands.

**1. Module-level `setInterval` runs on import.** As written, importing `mcp.ts` from any consumer (tests, other handlers, ports) starts the cleanup timer. `.unref()` is called so it won't block process exit, but it does make the module import-order-sensitive and adds a global side effect. Consider extracting to a `startSessionCleanup()` function that `index.ts` invokes once at boot. Keeps the module side-effect-free and gives tests a hook to opt in.

**2. The cleanup test doesn't invoke the actual cleanup path.** The third test manually deletes from both maps and asserts they're gone, which tests `delete` rather than the setInterval callback. Fake timers would pin the real behaviour:

```ts
jest.useFakeTimers();
sessionLastActivity[oldId] = Date.now() - SESSION_TIMEOUT_MS - 1;
sessionLastActivity[freshId] = Date.now();
jest.advanceTimersByTime(CLEANUP_INTERVAL_MS);
expect(transports[oldId]).toBeUndefined();
expect(transports[freshId]).toBeDefined();
```

Covers "idle removed" and "fresh preserved" in one pass.

**3. Cleanup deletes the reference but doesn't `.close()` the transport.** For SSE transports especially, the response stream may still be open server-side. Worth calling `transport.close?.()` before the `delete` so any pending resources actually get released.

**Smaller notes:**

- `console.error` for `session_cleanup` reads as loud. The event is expected behaviour, not an error. `console.log` matches `session_closed` and keeps ops dashboards quieter.
- `SESSION_TIMEOUT_MS` and `CLEANUP_INTERVAL_MS` as env-configurable would let production tune the values without a code change.
- Exporting `transports` opens internal state to any caller. If it's exported purely for testing, worth a comment saying so, otherwise other handlers may start reaching in.

Direction is right and the fix is real. Feel free to push back on any of these.

Satvik77777:
@JayDS22 — addressed all your review points:

Extracted the module-level setInterval into startSessionCleanup(), now called once from index.ts at boot — no more side effect on import
Rewrote the cleanup test using jest.useFakeTimers() + jest.advanceTimersByTime() to test the real interval callback, covering both idle removal and fresh-session preservation in one test
Added transport.close?.() in the cleanup interval's forced-removal path. Note: I initially added it to the onclose/SSE res.on("close") handlers too, but caught that this could cause recursive calls (checked the SDK source — close() isn't idempotent and triggers onclose again), so I removed it from those two paths and kept it only where it's actually needed
Changed console.error → console.log for the routine session_cleanup event
Made SESSION_TIMEOUT_MS/CLEANUP_INTERVAL_MS env-configurable with a proper isNaN-based parser (so 0 is respected, not silently overridden)
Added a comment explaining transports is exported for testing

Also synced with main (branch was quite behind — merged in #196/#199/#200/#201/#202/#208's changes). 26/26 tests passing.

JayDS22 (approved):
Solid revision. Fake timer tests pin the actual cleanup path, `startSessionCleanup()` from `index.ts` is the right decomposition, and the recursion catch on `close?.()` for the two `onclose` paths shows real thought. The idempotency guard around `clearInterval` and the `parseEnvMs` respecting `0` are nice touches beyond what I asked for.

LGTM.

PR #195, opened 2026-06-13
fix: reslate docs build on Windows to resolve spawn EINVAL

mttrbrts:
## Summary

Fixes two Windows CI failures. The first was pre-existing on `main`; the second was masked by the first and exposed once it was fixed.

### Fix 1 — `spawn EINVAL` in reslate (build step)

`reslate.js` calls `cp.spawn('npm.cmd', ['run', 'build.local'], { stdio: 'inherit', ... })` without `shell: true`. Node.js 20+ requires `shell: true` when spawning `.cmd` files on Windows; without it Node throws `EINVAL`.

Upgrading reslate isn't an option — there are only 6 pre-releases on npm (all `3.0.0-x`) and the current upstream `main` of [Mermade/reslate](https://github.com/Mermade/reslate) has the same bug with no fix in sight.

**Fix:** add `build-reslate.js`, a thin wrapper that replicates what reslate does internally (setting `SLATEDIR`/`NODE_PATH` env vars, then spawning `npm run build.local` in the reslate package dir) but with `shell: true`. The `build` script now calls `node build-reslate.js` instead of `npx reslate build`. This fixes both local Windows dev and CI without bypassing or skipping the docs build.

### Fix 2 — PowerShell `$LASTEXITCODE` not reset by `echo` (test step)

`npm test || echo "No tests configured"` works in bash (Linux/macOS) but fails on Windows because GitHub Actions uses PowerShell there. In PowerShell, `echo` is an alias for `Write-Output` — a PS cmdlet, not an external command — so `$LASTEXITCODE` retains the exit code from the failed `npm test` call even after the `||` branch runs. GitHub Actions uses `$LASTEXITCODE` to determine the step exit code, so the step exits 1 even though the fallback echo ran successfully.

**Fix:** add `shell: bash` to the test step. Git bash is always available on GH Actions Windows runners and makes `||` behave identically to Linux/macOS.

## Test plan

- [ ] Windows CI matrix (18.x and 20.x) — build and test steps both pass
- [ ] Linux/macOS CI matrix — full `build` (including reslate docs) still works
- [ ] `npm run build` on a local Windows machine builds the Slate docs site correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,8 +16,8 @@ jobs:
     strategy:
       matrix:
         node-version:
-          - 18.x
           - 20.x
+          - 22.x
         os:
           - windows-latest
           - macos-latest
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,12 +38,7 @@ jobs:
         run: npm install -g lockfile-lint syncpack
 
       - run: npm ci
-      - name: Build (with docs)
-        if: runner.os != 'Windows'
-        run: npm run build --if-present
-      - name: Build (skip reslate docs on Windows — spawn EINVAL with .cmd files)
-        if: runner.os == 'Windows'
-        run: npm run build:nodocs --if-present
+      - run: npm run build --if-present
       - name: Run tests
         shell: bash
         env:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,7 +38,12 @@ jobs:
         run: npm install -g lockfile-lint syncpack
 
       - run: npm ci
-      - run: npm run build --if-present
+      - name: Build (with docs)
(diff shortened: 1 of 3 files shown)


JayDS22:
Hi @mttrbrts, this PR is blocking four open GSoC PRs (#196, #197, #198, #199), all failing on the same `spawn EINVAL` in the build step that this fix addresses. Any chance of a review pass this week?

mttrbrts:
Here you go @JayDS22!

JayDS22:
> Here you go @JayDS22!

Thanks a lot, @mttrbrts Really appreciate the quick turnaround. This unblocks the four dependent GSoC PRs, so it's a huge help. I'll rebase them onto `main` and get them moving. Thanks again!

PR #197, opened 2026-06-14
feat(server): slice 2 typed errors (agreements + templatebuilder)

JayDS22:
Follow-up to #184 per @niallroche's "two updates per PR" guidance and @mttrbrts's "the other handlers can come later" approval comment.

Converts the remaining bare `throw new Error()` patterns in the agreements flow to typed `ServiceError` subclasses. Marked draft because slice 3 (mcp.ts) will land separately once @Satvik77777's overlapping PRs (#189, #192, #194) settle.

## What's wired

**`agreements.ts`** (`resolveAgreement`):
- "Cached template missing from database" → `TemplateNotFoundError` with `cached hash {hash}` as the identifier
- "Template with URI X referenced by agreement Y does not exist" → `TemplateNotFoundError(templateUri)`

A missing template now surfaces as **404 with structured body** instead of a bare 500.

**`agreements.ts`** (trigger handler):
- "Invalid request type: X. Expected one of: Y" → `InvalidPayloadError` with `received` and `expected` in details

MCP clients can now distinguish a bad `$class` from other failure modes.

**`templatebuilder.ts`** (`templateFromDatabase`):
- "Model type is not supported" → `InvalidPayloadError` with `received` and `expected` `$class` in details

This is the helper every agreement-resolution path calls; the typed error propagates to whichever route triggered it.

## Demo

Before:
```
POST /agreements/1/trigger (template URI broken)
HTTP 500  {"error":"Template with uri test://... referenced by agreement 1 does not exist"}
```

After:
```
POST /agreements/1/trigger (template URI broken)
HTTP 404  {
  "error": {
    "code": "TEMPLATE_NOT_FOUND",
    "message": "Template not found: test://...",
    "details": { "identifier": "test://..." }
  }
}
```

## Tests

- `agreements.test.ts`: the existing "should return 500 when template referenced by agreement does not exist" test is updated to assert the new 404 + structured body (that test was itself encoding the old bug, same pattern as the slice 1 update for "agreement not found").
- Full server suite: **64/64 passing**.

## Out of scope (slice 3)

The eight `throw new Error(errorMsg)` calls in `mcp.ts` are deliberately left for slice 3, since `mcp.ts` is also touched by @Satvik77777's open #189, #192, and #194. Once those merge (or the architecture in #192 settles), slice 3 will wire mcp.ts to ServiceError on top of whatever shape mcp.ts ends up in.

cc @mttrbrts @niallroche

### DIFF
--- a/server/handlers/agreements.test.ts
+++ b/server/handlers/agreements.test.ts
@@ -324,7 +324,7 @@ describe('Agreements Router - POST /:id/trigger', () => {
             });
         });
 
-        it('should return 500 when template referenced by agreement does not exist', async () => {
+        it('should return 404 with TEMPLATE_NOT_FOUND when template referenced by agreement does not exist', async () => {
             const mockAgreementData = {
                 id: 1,
                 template: 'test://template/nonexistent'
@@ -338,10 +338,14 @@ describe('Agreements Router - POST /:id/trigger', () => {
             const response = await request(app)
                 .post('/agreements/1/trigger')
                 .send({})
-                .expect(500);
+                .expect(404);
 
             expect(response.body).toEqual({
-                error: 'Template with uri test://template/nonexistent referenced by agreement 1 does not exist'
+                error: {
+                    code: 'TEMPLATE_NOT_FOUND',
+                    message: 'Template not found: test://template/nonexistent',
+                    details: { identifier: 'test://template/nonexistent' },
+                },
             });
         });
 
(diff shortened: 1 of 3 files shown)


JayDS22:
Closing this draft. Slice-2 typed-error work (agreements + templatebuilder) is still on the roadmap and tracked in PR #198, but holding the branch open for 10 days without commits is signalling stale work rather than coordination. Will re-open as a fresh PR after slice-3 (#200) clears its rework items, so the typed-error rollout merges in a clean order without four PRs racing for the same `server/handlers/mcp.test.ts` create-new diff. No code or commits lost; the branch is at `fork/jay/w3/typed-errors-slice-2` for reference.

PR #198, opened 2026-06-14
docs(gsoc-2026-roadmap): W3+W4 refresh

JayDS22:
Follow-up to merged #190 with the W3 day-1 status. Lightweight content-only update; no structural changes to the doc.

## What's updated

- **Status line** to Jun 15, 2026 / W3 day 1
- **Workstream 1** marked active with slice 1 merged (#184), slice 2 draft (#197), and tests follow-up (#196)
- **Schedule table**: W1 + W2 marked done; W3 marked active
- **Decisions table**: Decision 2 (alternatives depth) now fully resolved with both spikes shipped and memo drafted
- **Three new pending decisions** (rows 6-8) surfaced from the W2 alternatives memo: MCP Tasks framing for agreement triggers, function-calling adapter codegen location, whether to run CrewAI as a third spike before W9
- **Contributions table**: #155, #184, #190 marked merged; #153 marked closed-per-plan; #196 and #197 added
- **W1-W2 spike deliverables**: LangGraph branch and decision memo branch now linked alongside the OpenAI fn-calling branch

## Not in scope of this PR

- Any roadmap restructure or workstream addition (that would warrant a discussion before opening)
- Updates to the references section (handled in the prior follow-up that added the AAIF article)

cc @niallroche @mttrbrts @sanketshevkar @dselman

PR #199, opened 2026-06-17
feat(mcp): Concerto typed-context hint + protocol.cto schema resource

### LINKED ISSUE #185: Discuss: APAP's Concerto-typed schemas give it GraphQL-shaped semantics, not just MCP-shaped
Filing this as a discussion issue per a topic Niall raised on the May 20 GSoC sync. Opening it here for visibility rather than as a design proposal.

## The observation

APAP's protocol (`model/protocol.cto` and the generated `openapi.json`) is more strongly typed than a typical MCP tool surface. Every resource type carries a Concerto `$class` discriminator:

```
org.accordproject.protocol@1.0.0.Template
org.accordproject.protocol@1.0.0.Agreement
org.accordproject.protocol@1.0.0.TemplateModel
org.accordproject.protocol@1.0.0.CtoModel
org.accordproject.protocol@1.0.0.Logic
```

That `$class` is functionally a runtime type tag, much like GraphQL's `__typename`. Concerto's inheritance (`asset Template extends ...`, transactions extending `Request`/`Response`) mirrors GraphQL's interfaces and unions. The relationships between resources (`Agreement -> Template -> TemplateModel -> CtoModel`) form a typed graph that today is materialized as nested JSON, but is fundamentally navigable.

By contrast, typical MCP tool schemas are opaque input/output JSON Schemas with no semantic relationship between tools. `getTemplate`, `getAgreement`, `triggerAgreement` are independent functions; nothing in the protocol surface says they operate on a shared type system.

## Why this matters

Several practical implications fall out of this shape:

1. **Typed client tooling**: APAP can already round-trip codegen (Concerto -> OpenAPI -> TS client). A GraphQL-style schema would let other languages target the same type graph without the OpenAPI intermediate.
2. **LLM context efficiency**: a typed query surface allows callers to request only the fields they need. Today an MCP `read_resource` on an agreement returns the whole row including `data`, `state`, `signatures`, `historyEntries`, etc. A typed selector lets the model ask for `agreement { id, agreementStatus, template { displayName } }` and skip the rest.
3. **Auto-tooling (Niall's "context window discipline" point)**: each Concerto class could expose its own narrowly-scoped tool with typed arguments, rather than a single tool taking a freeform JSON blob. The model only loads the tools it needs.
4. **Privacy boundary (Niall's "scoped data per tool" point)**: per-field access control is natural when fields are typed and named. With JSON blobs you have to filter at the serialization layer.
5. **Federation**: APAP could expose itself as a GraphQL endpoint with minimal extra work, sitting alongside the existing REST and MCP surfaces.

## What I'm not proposing

I'm not proposing replacing REST or MCP, or shipping a GraphQL endpoint as a GSoC deliverable. The point of the issue is to make the type-system shape an explicit thing the project acknowledges, so design choices around auto-tooling, privacy scoping, and the post-2026-07-28 MCP spec compatibility lane can lean into it rather than treat the protocol as opaque blobs.

## Suggested discussion

- Does this framing land for other maintainers?
- Are there specific surfaces (auto-tooling, scoped reads, per-field permissions) where leaning into the typed shape would unblock work?
- Is there appetite for a GraphQL surface as a future workstream, or is "MCP + REST" the bounded scope?

Happy to do follow-up work on any of (1)-(4) above depending on direction.

JayDS22:
## Summary

Closes the discussion thread on #185 by implementing the smaller of the two shapes Dan and I converged on: a one-line Concerto-context hint via `InitializeResult.instructions` plus serving the protocol model itself as a readable MCP resource at `apap://schema/protocol.cto`.

No protocol changes. No transport changes. No service-layer changes. The MCP tool set and routes are untouched.

```diff
- new McpServer({ name: ..., version: ... }, { capabilities: { logging: {} } });
+ new McpServer({ name: ..., version: ... }, { capabilities: { logging: {} }, instructions: SERVER_INSTRUCTIONS });
```

Plus one new `server.resource(...)` registration that serves `model/protocol.cto` as `text/x-concerto`.

## Why

From the #185 discussion: an MCP client (and any LLM behind it) sees response bodies with `$class` discriminators like `org.accordproject.protocol@1.0.0.Template` but has no protocol-level cue that these are Concerto-serialized. Dan's point on the issue was that foundation models already have good Concerto knowledge — they just need to be told to use it. This change does exactly that, without inventing a new typed-selector layer.

## Empirical motivation

A directional first-pass A/B against the prototype (visible upstream as JayDS22/apap-mcp-poc#3) on a fixed 10-query set at temperature 0 showed both Anthropic Sonnet 4.6 and OpenAI gpt-4o consistently preferring the treatment over the control, with gpt-4o showing a larger lift, consistent with Sonnet having more Concerto knowledge already internalised.

**The exact percentage numbers initially shared with mentors are being withheld from this PR body** because the first-pass harness used single-run-per-arm scoring with substring keyword matching, which is not statistically defensible on its own. A multi-run + paired-statistics rerun with a stricter rubric is scheduled for W6 (see roadmap PR #198 and the rigor fix already shipped in POC PR #3). Until that rerun publishes, the qualitative findings below are the load-bearing evidence for this PR, and the change itself is small enough to evaluate on its own merits.

Two qualitative findings from the run:

1. **Tool routing improves on the OpenAI side.** On "Convert agreement 1 to HTML", gpt-4o control made no tool call and asked the user for the agreement ID despite it being in the prompt. Treatment called `convert-agreement-to-format` cleanly.
2. **Wrapper-vs-nested `\$class` is the one remaining gap.** Asked for an agreement's `\$class`, both gpt-4o variants reported the *inner* `data.\$class` rather than the outer `org.accordproject.protocol@1.0.0.Agreement`. Claude treatment got this right. Worth noting as future-iteration material; not a blocker for this PR.

## Implementation notes

- `SERVER_INSTRUCTIONS` is a static constant. Single-source-of-truth for the typed-context wording.
- `loadProtocolCto()` is dual-path: resolves `model/protocol.cto` whether running from `src/` (jest) or `dist/` (node). The two `path.resolve` candidates are tried in order; first hit wins. Cached after first read so the resource handler is filesystem-free on the hot path.
- The resource registration uses the same `server.resource(name, uri, callback)` shape as the existing template/agreement registrations for consistency. SDK-deprecation warnings on that signature affect the whole file equally; migrating to the newer SDK signature is separate work.

## Test plan

- [x] `npm test` passes — 67/67 (was 64; +3 new tests in `handlers/mcp.test.ts`)
- [x] `npm run build` produces `dist/handlers/mcp.js`; runtime path resolution to `model/protocol.cto` verified by `node -e \"require('./dist/handlers/mcp').loadProtocolCto()\"`
- [x] Jest test exercises the source path; production build exercises the dist path
- [x] End-to-end MCP client verification against `docker compose up`: `Client.getInstructions()` returns the typed-context string (357 chars), `apap://schema/protocol.cto` lists under resources, `readResource` returns the model bytes (7202 chars, `text/x-concerto`, namespace matches). This check caught a Dockerfile gap (`model/` was outside the build context), now fixed in this PR's second commit.

## Related

- #185 — the discussion that motivated this
- Prototype PRs on the POC repo (for reference and the A/B harness): https://github.com/JayDS22/apap-mcp-poc/pull/2 and https://github.com/JayDS22/apap-mcp-poc/pull/3

Closes #185

### DIFF
--- a/server/handlers/mcp.test.ts
+++ b/server/handlers/mcp.test.ts
@@ -3,6 +3,8 @@ import {
     serviceErrorToCallToolResult,
     serviceErrorToResourceError,
     buildApiErrorMessage,
+    PROTOCOL_CTO,
+    SERVER_INSTRUCTIONS,
 } from './mcp';
 import { McpError, ErrorCode } from '@modelcontextprotocol/sdk/types.js';
 import {
@@ -207,6 +209,22 @@ describe('mcp typed error helpers', () => {
             expect(data.error.code).toBe('CUSTOM');
             expect(data.error.message).toBe('I am a teapot');
             expect(data.error.details).toEqual({ teapot: true });
+        });
+    });
+});
+
+describe('Concerto typed-context (instructions + schema resource)', () => {
+    describe('SERVER_INSTRUCTIONS', () => {
+        it('mentions Concerto and the schema resource URI', () => {
+            expect(SERVER_INSTRUCTIONS).toMatch(/Concerto/);
+            expect(SERVER_INSTRUCTIONS).toMatch(/apap:\/\/schema\/protocol\.cto/);
+            expect(SERVER_INSTRUCTIONS).toMatch(/\$class/);
+        });
+    });
+
+    describe('PROTOCOL_CTO', () => {
+        it('decodes the embedded MODEL constant to the Concerto source', () => {
+            expect(PROTOCOL_CTO).toContain('namespace org.accordproject.protocol');
         });
     });
 });
--- a/server/handlers/mcp.ts
+++ b/server/handlers/mcp.ts
@@ -6,7 +6,7 @@ import { StreamableHTTPServerTransport } from "@modelcontextprotocol/sdk/server/
 import { isInitializeRequest, CallToolResult, GetPromptResult, ReadResourceResult, McpError, ErrorCode } from "@modelcontextprotocol/sdk/types.js";
 import * as crypto from "crypto";
 import { InMemoryEventStore } from './inmemoryeventstore';
-import { Agreement, Template } from '../db/schema';
+import { Agreement, MODEL, Template } from '../db/schema';
 import {
     ServiceError,
     TemplateNotFoundError,
@@ -22,6 +22,28 @@ const API_BASE_URL = process.env.API_BASE_URL || `http://${HOST}:${PORT}`
 
 // Get API authorization header from environment variable (optional)
 const API_AUTH_HEADER = process.env.APAP_API_AUTH_HEADER;
+
+// Concerto typed-context hint, exposed via MCP `InitializeResult.instructions`
+// and as a readable schema resource. Tells the client (and any LLM behind it)
+// that response payloads are Concerto-serialized so `$class` discriminators
+// can be interpreted directly against the protocol model.
+//
+// See accordproject/apap#185 for the discussion that motivated this. The
+// empirical A/B that produced this came out at Sonnet 4.6 +0.200 / gpt-4o
+// +0.383 mean score on a fixed query set.
+export const SERVER_INSTRUCTIONS = [
+    'Responses from this server are Concerto-serialized objects from the Accord',
+    'Project Agreement Protocol (APAP). Each resource carries a `$class`',
+    'discriminator (e.g. "org.accordproject.protocol@1.0.0.Template") identifying',
+    'its type and inheritance. The canonical Concerto model is available at',
+    '`apap://schema/protocol.cto` and can be read for type definitions.',
+].join(' ');
+
+// The Concerto model is embedded in db/schema.ts as a base64 constant at
+// drizzle-gen time (same source as `handlers/concertovalidation`), so no
+// filesystem access is needed at runtime and no build-time file copy is
+// required to serve the `apap://schema/protocol.cto` resource.
+export const PROTOCOL_CTO = Buffer.from(MODEL, 'base64').toString('utf-8');
 
 /**
  * @param url The APAP REST endpoint to call.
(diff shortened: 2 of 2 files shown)


niallroche (commented):
this is good overall but there is an issue 

compose.yaml now sets context to the repo root, instead of the server dir. Anyone running docker build directly from server/ will get a build failure. There is a logical separation between server and other parts of the repo and this would cause other side effects. Adding the entire repo root to the Docker daemon means client/, docs/, slate/, _site/, and everything else goes in there too.

I suggest either moving the model file into into server/model/ or possibly adding a root .dockerignore

JayDS22:
Different fix in `392c27b` after your audit, @niallroche — swapped the disk-load for the base64 `MODEL` constant that `drizzle-gen` already emits into `db/schema.ts`, which let me revert the compose and Dockerfile changes entirely.

1. **`apap://schema/protocol.cto` now reads from `Buffer.from(MODEL, 'base64').toString('utf-8')`.** Same source that `handlers/concertovalidation` already consumes, so no new copy step and no runtime `fs.readFileSync`. The `loadProtocolCto` function and its two `path.resolve` candidates are gone.
2. **`server/compose.yaml` back to `context: .`, `server/Dockerfile` back to `COPY . .`.** No repo-root Docker context, `docker build` from `server/` works the same way it did before this PR touched it.

Net diff against main is 2 files now (`handlers/mcp.ts` + `handlers/mcp.test.ts`, +57/-2). Rebased on top of #200 so the typed-error changes are in the base.

Verification: `npx tsc --noEmit -p .` clean, `npm test` 85/85 pass (16 in `mcp.test.ts` including the 2 new Concerto-context tests). CI matrix green.

Marked ready for review.

niallroche:
The Docker section in #202 is now incorrect. Suggest you update #202 before or alongside merging #199, otherwise the CLAUDE.md will document a change that doesn't exist.

Also $class lacks backtick formatting in SERVER_INSTRUCTIONS

JayDS22:
Pushed `9d48a90`, @niallroche. Reading the audit as covering both code-like tokens in the string: `` `$class` `` already had backticks in the source but the immediately-following type example `"org.accordproject.protocol@1.0.0.Template"` was in escaped double quotes, so wrapped that in backticks too for consistency.

Rendered instructions now read:

> ...Each resource carries a `` `$class` `` discriminator (e.g. `` `org.accordproject.protocol@1.0.0.Template` ``) identifying its type and inheritance. The canonical Concerto model is available at `` `apap://schema/protocol.cto` `` and can be read for type definitions.

Separately on the #202 stale-doc call-out — will land a rewrite there in a follow-up push, aiming to have it in before this merges so the CLAUDE.md doesn't describe a change that no longer exists.

niallroche (approved):
issues have been addressed

Issue #143, opened 2026-03-26
[RFC] Refactor MCP tools to call shared service layer directly


Thomas-Sedhom:
This issue for discussion.

Currently, the MCP server in handlers/mcp.ts calls the REST API over an internal HTTP loop via makeApiRequest() just to reuse logic that already exists in the same process. This proposal suggests extracting that logic into a shared service layer that both the REST handlers and the MCP tools can call directly.

Current Problem
Looking at handlers/mcp.ts, every tool makes an HTTP call back to the same server:
```
const result = await makeApiRequest(`${API_BASE_URL}/agreements/${agreementId}`);
```

This means an MCP tool call makes a full network round-trip within the same process — adding unnecessary latency, making the code harder to test in isolation, and creating a hidden dependency on the server being reachable at `API_BASE_URL`.

## Proposed Solution

Extract three service classes:
- `TemplateService`
- `AgreementService`  
- `SharedModelService`

Both the REST handlers and the MCP tools would call these services directly:

<img width="1010" height="634" alt="Image" src="https://github.com/user-attachments/assets/16233c5f-a032-4487-b4cb-4a8501cc9355" />

This approach:
- Keeps the REST API fully intact and accessible to any external client
- Eliminates the internal HTTP round-trip from MCP tools
- Makes both layers independently testable
- Creates a clean separation of concerns

JayDS22:
Strong proposal @Thomas-Sedhom. The `makeApiRequest()` loop is exactly the kind of hidden coupling that breaks down at scale. Fully agree it needs to go.

The three-service split makes sense. One thing I'd add: the service layer should own validation, not just business logic. #136 showed PUT routes bypassing validation that POST enforces. If both REST handlers and MCP tools call the same service methods that validate before executing, you eliminate that entire class of bugs by construction.

Practical concern on migration: swapping out `makeApiRequest()` across all MCP tools in one PR is risky. Safer to extract one service at a time (TemplateService first since it's the most exercised path), verify REST and MCP still behave identically, then repeat. Each PR stays reviewable and reversible.

JayDS22:
Proposing two `ServiceError` subclasses to land alongside the slice-3 typed-errors PR (#200), so the three `// TODO: #143` sites in `server/handlers/mcp.ts` have a home in the typed hierarchy rather than continuing to throw bare `Error`:

```ts
// server/services/errors.ts

export class UpstreamApiError extends ServiceError {
  constructor(
    public readonly upstreamUrl: string,
[...]

PR #200, opened 2026-06-25
feat(server): typed errors in mcp.ts (slice 3)

JayDS22:
## Summary

Completes the typed-error rollout started in #184 (slice 1, infrastructure + `agreements.ts` handler) and continued in #197 (slice 2) by migrating the eight bare `throw new Error(...)` sites in `server/handlers/mcp.ts` to the existing `ServiceError` hierarchy in `server/services/errors.ts`. MCP is the protocol surface the typed-error contract was always meant to serve: MCP clients now receive structured `{ code, message, details }` payloads instead of opaque SDK error strings on 404s and conversion failures.

## Throw-site mapping

| Site | New behavior |
| --- | --- |
| `getAgreement` resource | 404 -> `AgreementNotFoundError` (wrapped for resource path) |
| `getTemplates` collection | bare `Error` + TODO (needs `UpstreamApiError`, see #143) |
| `getAgreements` collection | bare `Error` + TODO (needs `UpstreamApiError`, see #143) |
| `draftAgreement` convert | 404 -> `AgreementNotFoundError`, else `AgreementConversionError` |
| `triggerAgreement` | 404 -> `AgreementNotFoundError`, else bare `Error` + TODO (needs `AgreementTriggerError`) |
| Template-by-ID resource | 404 -> `TemplateNotFoundError` (wrapped for resource path) |
| `getTemplate` tool | 404 -> `CallToolResult { isError: true, content: [...] }` carrying `TemplateNotFoundError.toJSON()` |
| `getAgreement` tool | 404 -> `CallToolResult { isError: true, content: [...] }` carrying `AgreementNotFoundError.toJSON()` |

## Why this is two helper functions, not one

MCP has two error surfaces and they are not symmetric:

* **Tool callbacks** can return `CallToolResult { isError: true, content }`, which is the spec-blessed channel for tool errors. `serviceErrorToCallToolResult` wraps a `ServiceError` into that shape with a single `text` content block whose body is `JSON.stringify(error.toJSON())`.
* **Resource handlers** have no equivalent `isError` channel. The SDK only propagates a thrown `Error.message` back to the client. `serviceErrorToResourceError` builds an `Error` whose `message` is the same JSON payload, so clients can `JSON.parse` it and branch on the machine-readable `code` exactly like REST callers already do.

This is intentionally simple. A spec-level `isError` channel for resource reads would be a better long-term answer but is out of scope for this slice.

## Why some sites stayed bare

The collection-fetch failures (`getTemplates`, `getAgreements`, list operations) and the non-404 trigger path do not map cleanly to any existing `ServiceError` subclass. Inventing `UpstreamApiError` / `AgreementTriggerError` in this PR would expand the hierarchy without an RFC discussion, so those sites are flagged with `// TODO: needs new ServiceError subclass ... discuss in #143 RFC` and left as bare `Error` throws. That keeps this slice scoped to the existing typed-error contract.

## Test plan

- [x] `npm test` from `server/`: 71 tests pass across 7 suites (previously 64 across 6).
- [x] `npx tsc -p . --noEmit` clean.
- [x] New `handlers/mcp.test.ts` exercises `serviceErrorToCallToolResult` against `TemplateNotFoundError`, `AgreementNotFoundError`, `AgreementConversionError`; `serviceErrorToResourceError` against typed subclasses and arbitrary `ServiceError`; and `buildApiErrorMessage` for body-readable and body-locked responses.
- [ ] Smoke test through `mcp-inspector` against a live RI (deferred to reviewer if requested).

## Related

* #184 - slice 1, introduces the `ServiceError` hierarchy and wires `agreements.ts`
* #197 - slice 2
* #143 - typed-error RFC tracking issue for the additional subclasses called out in the TODOs above

### DIFF
--- a/server/handlers/mcp.test.ts
+++ b/server/handlers/mcp.test.ts
@@ -1,4 +1,13 @@
-import { buildApiErrorMessage } from './mcp';
+import { jest } from '@jest/globals';
+import { __testables, buildApiErrorMessage } from './mcp';
+import {
+    AgreementConversionError,
+    AgreementNotFoundError,
+    ServiceError,
+    TemplateNotFoundError,
+} from '../services/errors';
+
+const { serviceErrorToCallToolResult, serviceErrorToResourceError } = __testables;
 
 // Helper to create a mock failed Response. Only the fields buildApiErrorMessage
 // actually reads (ok, status, text) need to be present; the cast is safe because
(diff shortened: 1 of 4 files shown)


JayDS22:
Moving to draft pending audit follow-ups before merge: swapping `throw new Error(JSON.stringify(error.toJSON()))` for `McpError(code, message, error.toJSON())` on the resource-handler paths (the SDK ships `McpError` with a `data` field for exactly this case), dropping the `__testables` named export in favour of plain `export`, and either migrating `triggerAgreement`'s non-404 branch or punting both 404 + non-404 to keep the function internally consistent. Will re-mark ready once those land.

JayDS22:
Marking ready. Three audit items addressed in 8e9a012:

1. **Resource handlers now throw `McpError`.** `serviceErrorToResourceError` returns an `McpError` carrying the typed payload in its `data` field, with the JSON-RPC code chosen by status (`InvalidParams` for 404-style, `InternalError` otherwise). No more `JSON.stringify(...)` tunnelled through `Error.message`. Clients can read `error.code` + `error.data.error.{code,message,details}` directly off the SDK error.
2. **`__testables` is gone.** The three helpers (`serviceErrorToCallToolResult`, `serviceErrorToResourceError`, `buildApiErrorMessage`) are now direct named exports on `handlers/mcp.ts` and the test file imports them by name. The testing-only object that hid this is removed.
3. **`triggerAgreement` non-404, `getTemplates`, `getAgreements` migrated.** All three remaining bare-throw sites now use typed subclasses. Added `UpstreamApiError` (502, carries upstream URL + HTTP status + raw body) and `AgreementTriggerError` (502, carries agreement id + upstream message) to `services/errors.ts`, matching the shape proposed in #143. The three `// TODO ... discuss in #143 RFC` comments are deleted. Tool callbacks for `getTemplate` / `getAgreement` also surface `UpstreamApiError` through `CallToolResult { isError: true }` rather than throwing through the SDK.

Tests now 77 pass across 7 suites (was 71). `npx tsc --noEmit -p .` clean. Branch rebased onto current main (picked up #195).

@niallroche ready for another look when you have a moment.

niallroche (approved):
buildApiErrorMessage is still called by name in draftAgreement but inlined in the resource handlers. This was already flagged this and Jay is deferring this it to a follow-up

PR #202, opened 2026-06-25
docs(claude-md): document MCP typed-context + Docker context for Claude Code

JayDS22:
## Summary

Splits the `CLAUDE.md` addition out of #199 (Concerto typed-context) into a trivial follow-up so #199's diff is feature-only.

`CLAUDE.md` is the contributor guidance file that Claude Code reads when working in this repo. Two recent changes earned a section:

1. The MCP server now advertises a Concerto typed-context via `InitializeResult.instructions` + a readable `apap://schema/protocol.cto` resource (#199). Document `loadProtocolCto()`'s dual-path probe and where the constants live so a future contributor doesn't reinvent the loader.
2. The Docker build context moved from `server/` to the repo root so `model/protocol.cto` can be bundled at runtime for the schema resource. A bare `docker build .` from inside `server/` now fails; the doc explains the corrected invocations.

## Why split

Original commit was on `jay/feat/mcp-concerto-context` (PR #199). Reviewers reading #199 should see feature + Docker context restructure only; a 95-line documentation change is orthogonal scope. Landing this separately keeps both PRs cheap to review.

## Test plan

- [x] Doc reads cleanly top-to-bottom
- [ ] No code changes; CI green is the only signal needed


niallroche:
I think there are three sections need updating before this merges:

As referenced in 199, the Docker build context paragraph — describes the context: .. change that was reverted in commit https://github.com/accordproject/apap/commit/392c27bbece2a31c0524e700391664a21be6dea5. The build context is still server/; this whole paragraph should be removed.

Typed-context surface → loadProtocolCto() — the function no longer exists. The final implementation uses export const PROTOCOL_CTO = Buffer.from(MODEL, 'base64').toString('utf-8') with no filesystem access. Rewrite to document the MODEL constant and the drizzle-gen dependency instead.

ServiceError status — "handlers/agreements.ts is the only handler currently wired" is stale since [#200](https://github.com/The-Building-Blocks/app-enterprise/issues/200) merged. mcp.ts is also wired now.

JayDS22:
Pushed `257b755`, @niallroche. Rebased on top of `6afe6a0` (post-#200) and reworked the three stale sections you called out.

1. **Docker build context paragraph removed.** The `context: ..` change was reverted in `392c27b` on #199, so the build context is still `server/` and `docker build .` from inside works normally. Nothing to document, so the whole paragraph is gone.
2. **`loadProtocolCto()` paragraph rewritten.** Now describes `PROTOCOL_CTO` as the runtime source, sourced from base64-decoding the `MODEL` constant that `scripts/drizzle-gen.ts` emits into `db/schema.ts` at build time. Same origin `handlers/concertovalidation` already consumes. Explicitly calls out that `npm run drizzle-gen` refreshes both the Drizzle schema and the embedded `MODEL` when `model/protocol.cto` changes.
3. **ServiceError rollout note updated.** `handlers/mcp.ts` is called out as the second wired handler post-#200 (uses `serviceErrorToCallToolResult` and `serviceErrorToResourceError`), so the "only handler currently wired" line no longer misrepresents the state on main.

Since section 2 references the `PROTOCOL_CTO` export from #199, this doc lands cleanly if it merges alongside or after #199 (per your earlier note about the timing).

niallroche (approved):

PR #203, opened 2026-06-25
fix(mcp): return mcp-session-id header on initialize

Satvik77777:
The mcp-session-id header was validated on incoming requests but never returned to clients on initialization, breaking clients that rely on session continuity per spec.

Split out from #189 per discussion with @niallroche.

### DIFF
--- a/server/handlers/mcp.test.ts
+++ b/server/handlers/mcp.test.ts
@@ -1,4 +1,6 @@
 import { jest } from '@jest/globals';
+import request from 'supertest';
+import express from 'express';
 import {
     serviceErrorToCallToolResult,
     serviceErrorToResourceError,
@@ -7,6 +9,7 @@ import {
     SERVER_INSTRUCTIONS,
     CACHE_HINTS,
 } from './mcp';
+import mcpRouter from './mcp';
 import { McpError, ErrorCode } from '@modelcontextprotocol/sdk/types.js';
 import {
     AgreementConversionError,
@@ -298,5 +301,35 @@ describe('SEP-2549 cache hints exposed by the MCP handler', () => {
             expect(Number.isFinite(hint.ttlMs)).toBe(true);
             expect(hint.ttlMs).toBeGreaterThan(0);
         }
+    });
+});
+
+describe('MCP HTTP Router', () => {
+    let app: express.Application;
+
+    beforeEach(() => {
+        app = express();
+        app.use(express.json());
+        app.use('/', mcpRouter);
+    });
+
+    it('returns mcp-session-id header on initialization response', async () => {
+        const res = await request(app)
+            .post('/mcp')
+            .set('Accept', 'application/json, text/event-stream')
+            .send({
+                jsonrpc: '2.0',
+                id: 1,
+                method: 'initialize',
+                params: {
+                    protocolVersion: '2025-03-26',
+                    capabilities: {},
+                    clientInfo: { name: 'test', version: '1.0' }
+                }
+            });
+        expect(res.status).toBe(200);
+        expect(res.headers['mcp-session-id']).toMatch(
+            /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/
+        );
     });
 });
--- a/server/handlers/mcp.ts
+++ b/server/handlers/mcp.ts
@@ -639,6 +639,7 @@ router.all('/mcp', async (req: Request, res: Response) => {
                     console.log(`StreamableHTTP session initialized with ID: ${sessionId}`);
                     transports[sessionId] = transport;
                     sessionLastActivity[sessionId] = Date.now();
+                    res.setHeader('mcp-session-id', sessionId);
                 }
             });
 


niallroche:
great @Satvik77777 this executes the split cleanly,  verification_report.md is committed to the repo root in this PR, can you please remove

Satvik77777:
Done — verification_report.md removed.

niallroche:
1. The rest of the codebase uses the structured logger (log from ../utils/logger) so use that instead of the console.log('Connected server to transport'); 

2. Header should be set inside onsessioninitialized, not outside

The current approach pre-generates the ID and passes a lambda that always returns it, then sets the header using the pre-generated value. This works coincidentally, but the SDK's sessionIdGenerator is a callback it controls — it could in principle call it multiple times or not at all before the session is established. The canonical session ID is sid in onsessioninitialized. The header should be set there:

onsessioninitialized: (sid) => {
  sessions[sid] = { transport, apiAuthHeader: authHeader };
  res.setHeader('mcp-session-id', sid);
}
This guarantees the header is only sent after the session actually exists, using the ID the SDK assigned — not a value the handler pre-computed before the SDK even ran.

3. Test file is in the wrong place

mcp-session-header.test.ts as a separate file conflicts with the existing mcp.test.ts that #196 and #200 already established. The session header test belongs in mcp.test.ts alongside the other MCP handler tests, not in a one-off file. Fragmented test files for the same handler make the suite harder to navigate.

4. Test uses Supertest against StreamableHTTP, which #189 specifically warned against

From #189's description: "StreamableHTTPServerTransport requires specific Accept headers and stream mechanics that supertest cannot emulate." This test sends Accept: application/json, text/event-stream and may be passing for the right reasons or not — it's worth confirming the isInitializeRequest branch is actually being reached before treating the test as reliable coverage.

Satvik77777:
@niallroche — following up on the logger note. Did a full search (file names, server/utils/, package.json deps, git history, and also checked #192 which centralizes error handling but still uses raw console.log/console.error) — there's no logger/log module anywhere in the repo, past, present, or in-flight across any open PR. Current convention: console.log (general), console.error (errors), console.info only in index.ts for lifecycle events, plus morgan for HTTP request logging (not a general logger). Did you mean something landing separately, or should I just use console.log/create a minimal logger as part of this fix? Whichever you prefer.

JayDS22:
@Satvik77777 dropped a quick pass on this. The fix is well-scoped: pre-generating the session ID and threading it through both the transport's `sessionIdGenerator` and the response header is the right shape for spec compliance. Two things worth flagging:

1. **`sessionIdGenerator` closure captures a single UUID.** Every call to that generator now returns the same `newSessionId` for the lifetime of this transport. That's fine for the typical initialize-once flow, but worth a comment explaining the intent so a future reader doesn't assume it's meant to generate multiple distinct IDs. Alternatively, since the SDK only calls it once per session, you could just pass the ID directly if the SDK API allows it.

2. **Test could catch more.** Current assertion is `toBeDefined()`, which passes even if the header is set to an empty string. Suggest strengthening to something like `expect(res.headers['mcp-session-id']).toMatch(/^[0-9a-f-]{36}$/i)` to pin down the UUID shape. A second test that reuses the returned session ID for a follow-up request would also prove the round-trip works end-to-end, not just that the header is written.

Header placement (line 461, before `transport.handleRequest` on line 477) looks correct. That call is where the response body gets written, so setting the header first is the right ordering.

Not a blocker from me. The split from #189 into #194 and this one reads cleanly.

Satvik77777:
Pushed the requested changes:

Session header now set inside onsessioninitialized using the SDK-assigned sid, per your suggestion @niallroche
Session header test moved into mcp.test.ts under describe('MCP HTTP Router'), with the assertion strengthened to a UUID format regex match
Confirmed via a temporary debug log (since removed) that the isInitializeRequest branch is genuinely exercised by the test, not passing incidentally

@JayDS22 — the sessionIdGenerator closure concern no longer applies: it's now sessionIdGenerator: () => crypto.randomUUID() with no pre-generated closure variable, and the canonical ID comes from onsessioninitialized. The UUID regex assertion is also already in place.
Still waiting on your input on the logger question above (@niallroche) before making that last change.

Satvik77777:
@JayDS22 — since you've touched mcp.ts extensively in #196/#200, quick check: did you ever come across a ../utils/logger module, even something planned/discussed elsewhere? Just following up since @niallroche's note referenced it.

JayDS22:
@Satvik77777 confirmed on my end too. Grep across `server/` (source, dist, and open branches), no `../utils/logger` or any `utils/` directory in the repo. Current convention is `console.log` / `console.error`, plus `morgan` for HTTP request logging. Nothing in the PRs I have touched (#184, #196, #200) or the ones I have open (#199, #201, #202) introduces a logger module.

One possible source of the reference: `apap-mcp-poc` (the POC that upstream is being brought toward) uses `pino` as its structured logger. Niall may have been thinking of the POC convention. Might be worth holding on that item until he clarifies which logger he wants introduced here, or whether it can defer to a follow-up PR.

niallroche:
ignore the comment about the logger @Satvik77777. Claude was hallucinating :-) as I was working on a different branch I started previously and it picked up on this incorrectly. I will merge when the conflicts are resolved

Satvik77777:
Conflicts resolved — merged latest upstream/main into this branch (picked up #201/#202 cleanly). All 23 tests passing, including the new SEP-2549/Concerto tests from main plus the session-header test. Ready for review whenever you get a chance

JayDS22 (approved):
Approving. The header is now set inside `onsessioninitialized` using the SDK-assigned `sessionId`, so `transports[sessionId]`, `sessionLastActivity[sessionId]`, and the response header all draw from the same source. That resolves the `sessionIdGenerator` closure concern from Jul 9. The generator is `() => crypto.randomUUID()` which produces a fresh UUID per invocation, so no single-UUID reuse.

Optional follow-up (not blocking): the regression test asserts UUID format on the response header but doesn't verify round-trip. A stronger version would send a second request with the returned sessionId and assert it reuses the same transport. Fine to defer.

niallroche (approved):
good work @Satvik77777

PR #126, opened 2026-03-17
test: Add Isolated Unit Tests for Dynamic HTTP Template Retrieval

rockaxorb13:
## Overview
This PR introduces robust, isolated unit tests for my previous PR #112  which added dynamic HTTP template retrieval pipeline in the APAP server. It ensures that when a user creates an agreement using an external HTTP template URI, the server correctly intercepts the URI, fetches the `.cta` archive, compiles it via Cicero, and stores the extracted models and logic in the database.

## Changes Included
* **`server/handlers/templatebuilder.test.ts` (New File):** * Created dedicated unit tests to verify the integrity of the template extraction pipeline (`extractTemplateForDatabase` and `templateFromDatabase`).
* **`server/handlers/agreements.test.ts` (Modified):** * Added a comprehensive integration test for the `POST /agreements/` route.
  * Implemented a context-aware `fetch` mock to simulate downloading `.cta` archives without hitting the external network, while allowing Cicero to fetch its required `.cto` dependencies normally.

## Technical Challenges Resolved
Testing this dynamic pipeline required disarming several complex Jest and Node.js environmental traps without altering a single line of production code:

1. **Jest Execution Timeouts:** Compiling Cicero templates heavily taxes single-core performance. The default 5000ms Jest timeout was frequently exceeded during local execution. Resolved by cleanly extending the timeout context (`jest.setTimeout(30000)`).
2. **Node.js Shared ArrayBuffer Pooling:** When mocking the `fetch` response to return the local template buffer, passing `templateBuffer.buffer` directly fed shared, unallocated memory garbage to Cicero, causing a silent 500 error. Resolved by cleanly slicing the exact byte coordinates from the memory pool (`buffer.slice(byteOffset, byteOffset + byteLength)`).
3. **Greedy Fetch Interceptors:** Our initial `fetch` mock successfully returned the `.cta` buffer but accidentally intercepted Cicero's internal engine attempting to download external `.cto` models, causing a `r.text is not a function` crash. Resolved by making the interceptor conditionally fallback to the `originalFetch` for non-archive requests.
4. **Module Override & Test Pollution:** Needed to bypass `concertoValidation` specifically for the creation route without breaking the older trigger tests. Resolved using `jest.requireActual` and surgical `.mockResolvedValueOnce()` injections on both default and named exports, preventing test-induced design damage to the production router.
[...]

PR #112, opened 2026-03-04
feat: implement build-time external template retrieval and hashing (#60)

rockaxorb13:
## Description

This PR addresses Issue #60 by providing a URI-based extensibility mechanism to pull APAP resources from externally managed registries. 

Based on maintainer feedback from PR #91, this implementation shifts the template fetching mechanism from *runtime* to **build-time** (during agreement creation). It ensures that the template logic is securely downloaded, hashed, and persisted directly into Postgres, eliminating the need to re-fetch or re-parse the template during subsequent contract triggers.

### 🎥 Video Demo
**[Watch the architecture and execution demo on Vimeo](https://vimeo.com/1168100968?share=copy&fl=sv&fe=ci)**

### 1. Pluggable Retriever Interface (`server/handlers/retrievers/`)
To support the long-term goal of pulling from various OCI registries, IPFS, or custom CMS tools, the retrieval logic has been fully abstracted:
* **`ITemplateRetriever.ts`**: A strict TypeScript interface defining how external archives should be fetched into memory.
* **`HttpTemplateRetriever.ts`**: The first plugin implementation, which securely handles standard `http://` and `https://` URIs.

### 2. Build-Time Fetching & Integrity Hashing (`server/handlers/agreements.ts`)
* **Agreement Creation (`POST /`)**: When an agreement is instantiated with an external template URI, the engine immediately delegates to the appropriate retriever plugin.
* **Unique Hashing**: The `.cta` archive is loaded into memory, and `apTemplate.getHash()` is utilized to calculate a unique hash that binds the template logic to the agreement.
* **Caching (`templatebuilder.ts`)**: The processed template logic is extracted and cached in the database alongside its unique hash.

### 3. Postgres Storage & Migrations (`server/db/schema.ts` & `server/drizzle/`)
* Updated the `Agreement` table to store the `templateHash`.
* Updated the `Template` table to cache the fully parsed Cicero template logic, preventing network latency and external dependencies during runtime execution.
* Generated new Drizzle migrations (`0001_simple_clint_barton.sql`) to reflect these schema changes.
* Regenerated OpenAPI specifications and TypeScript client definitions.

## Testing & Validation
* Added local `.cta` test archives (e.g., `late-delivery.cta`) to validate the buffer loading and database insertion mechanics.
* Verified that subsequent triggers properly bypass the external network and successfully utilize the cached template logic from Postgres.

**Closes:** #60 
**Replaces:** #91

PR #204, opened 2026-07-03
feat: implement HTTP retrieval for external shared models

### LINKED ISSUE #60: Support for external template registry/repository/OCI
Provide a URI-based extensibility mechanism, so that APAP resources can be pulled from externally managed registries/repositories.

This URI could link to something like:
- to a publicly hosted HTTPS location (we currently host at https://templates.accordproject.org/)
- that could point to internal systems like a CMS or CLM tool
- to a GitHub url
- to a decentralised file system like IPFS
- a container-like registry

We could look at relaunching legalschema.org / https://github.com/legalschema as a potential source we could help to administer

We could use OCI registries to store templates, models, and misc. related data. The [OCI Distribution Spec](https://github.com/opencontainers/distribution-spec) provides a generic protocol for using registries (the ones we use typically use for containers) as a storage system. Basically, we'd use registries the same way Helm charts do. Individual artifacts could be stored in the registries or packaged in `tar` format. 

[ORAS](https://oras.land/) (OCI Registry As Storage) is an open source sandbox project under the CNCF that facilitates this use case very simply. They give a very simple and easy to understand[ introduction here](https://oras.land/docs/).

Pursuing this allows us to take advantage of a mature, performant distribution mechanism with wide support, without needing to invest significant resources.

**Customizing for Accord**

We could enable custom tooling and analysis for our specific artifacts by labeling them with media types such as:

- `application/vnd.accord.template.data.v1+json`
- `application/vnd.accord.concerto.data.v1+json`
- `application/vnd.accord.concerto.model.v1+cto`

**Scenario**

Decombine might host a registry repository specifically for open source templates facilitating sales agreements. We would create a repository at `docker.io/decombine/sales-agreements`. We could then `push` individual templates to the `sales-agreement` repository. The `docker.io/decombine/sales-agreements:v0.0.5-en` could be added as a reference in APAP.

**For your review**

- [Distributing Supply Chain Artifacts with OCI & ORAS Artifacts](https://youtu.be/BpKF_0M37-0?si=WB8GgiZ8owAf9U5r)

rockaxorb13:
Hello @mttrbrts and @dselman, apologies for being missing in action, I had my end semester exams, followed by a vacation but I am back and plan to keep on contributing & learning here on! 

With this PR, I am continuing the work which I started with PRs #112 and #126

## Description

This PR introduces the ability to dynamically fetch and import external Shared Models (`.cto` files) via HTTP/HTTPS URIs, directly addressing the next phase of extensibility outlined in #60. 

This builds upon the pattern established in the external template retrieval feature, providing APAP with the capability to pull lightweight model dictionaries from external registries or public repositories prior to database insertion.

### Architectural Changes

* **Retriever Interfaces:** Established a new `IModelRetriever` interface to cleanly separate model-fetching logic from binary template-fetching logic (`ITemplateRetriever`).
* **HTTP Implementation:** Added `HttpModelRetriever` to securely handle network requests for external `.cto` text files, passing along `EXTERNAL_TEMPLATE_TOKEN` headers when present.
* **POST Route Interception:** Modified the `POST /` route in `server/handlers/sharedmodels.ts` to intercept requests containing external URIs.
* **Native Concerto Validation:** The intercepted raw `.cto` text is piped through `@accordproject/concerto-core`'s `ModelManager` to validate the syntax and extract the namespace.
* **Data Formatting:** Automatically wraps the validated external text into the `org.accordproject.protocol@1.0.0.CtoModel` JSON format expected by the APAP database schema, seamlessly handing off the payload to the existing `crudRouter`.

### Motivation and Context

As noted by @dselman in Issue #60, APAP requires extensibility for each core resource type (Agreements, Templates, and Shared Models). While previous PRs handled the binary `.cta` template archives, Shared Models require a distinct pipeline because they are fetched as raw text rather than zip buffers, and do not possess the embedded logic/grammar layers of a full template.

By parsing the external text through `ModelManager` prior to database insertion, we guarantee that only valid Concerto models enter the Postgres database, preventing corrupted states down the line. 

### Manual Verification Steps

1. Start the local APAP server.
2. Send a `POST` request to `/sharedmodels` with an external URI:
   ```json
   {
     "uri": "[https://models.accordproject.org/money@0.3.0.cto](https://models.accordproject.org/money@0.3.0.cto)"
   }


Fixes #60 (Partial - Extends external URI support to Shared Models)

### DIFF
--- a/server/handlers/retrievers/HttpModelRetriever.ts
+++ b/server/handlers/retrievers/HttpModelRetriever.ts
@@ -1,0 +1,27 @@
+import { IModelRetriever } from './IModelRetriever';
+
+export class HttpModelRetriever implements IModelRetriever {
+    public getURISchemes(): string[] {
+        return ['http', 'https'];
+    }
+
+    async fetchModel(uri: string): Promise<string> {
+        if (!uri.startsWith('http://') && !uri.startsWith('https://')) {
+            throw new Error(`Invalid URI scheme: ${uri}`);
+        }
+
+        const headers: Record<string, string> = {};
+
+        if (process.env.EXTERNAL_TEMPLATE_TOKEN) {
+            headers['Authorization'] = `Bearer ${process.env.EXTERNAL_TEMPLATE_TOKEN}`;
+        }
+
+        const response = await fetch(uri, { headers });
+        
+        if (!response.ok) {
+            throw new Error(`Failed to fetch model from ${uri}: ${response.statusText} (${response.status})`);
+        }
+
+        return await response.text();
+    }
+}
--- a/server/handlers/retrievers/HttpModelRetriever.ts
+++ b/server/handlers/retrievers/HttpModelRetriever.ts
@@ -1,13 +1,32 @@
 import { IModelRetriever } from './IModelRetriever';
 
+const ALLOWED_DOMAINS = [
+    'models.accordproject.org',
+    'templates.accordproject.org',
+    'raw.githubusercontent.com'
+];
+
 export class HttpModelRetriever implements IModelRetriever {
     public getURISchemes(): string[] {
         return ['http', 'https'];
     }
 
+    private isAllowedUrl(urlString: string): boolean {
+        try {
+            const parsedUrl = new URL(urlString);
+            return ALLOWED_DOMAINS.includes(parsedUrl.hostname);
+        } catch (e) {
+            return false;
+        }
+    }
+
     async fetchModel(uri: string): Promise<string> {
         if (!uri.startsWith('http://') && !uri.startsWith('https://')) {
             throw new Error(`Invalid URI scheme: ${uri}`);
+        }
+
+        if (!this.isAllowedUrl(uri)) {
+            throw new Error(`SSRF Prevention: Domain not in allowlist. Received: ${uri}`);
         }
 
         const headers: Record<string, string> = {};
@@ -16,12 +35,22 @@ export class HttpModelRetriever implements IModelRetriever {
             headers['Authorization'] = `Bearer ${process.env.EXTERNAL_TEMPLATE_TOKEN}`;
         }
 
-        const response = await fetch(uri, { headers });
-        
-        if (!response.ok) {
-            throw new Error(`Failed to fetch model from ${uri}: ${response.statusText} (${response.status})`);
-        }
+        const controller = new AbortController();
+        const timeoutId = setTimeout(() => controller.abort(), 5000);
 
-        return await response.text();
+        try {
+            const response = await fetch(uri, { 
+                headers,
+                signal: controller.signal
+            });
+            
+            if (!response.ok) {
+                throw new Error(`Failed to fetch model from ${uri}: ${response.statusText} (${response.status})`);
+            }
+
+            return await response.text();
+        } finally {
+            clearTimeout(timeoutId);
+        }
     }
 }
--- a/server/handlers/retrievers/HttpModelRetriever.ts
+++ b/server/handlers/retrievers/HttpModelRetriever.ts
@@ -11,26 +11,25 @@ export class HttpModelRetriever implements IModelRetriever {
         return ['http', 'https'];
     }
 
-    private isAllowedUrl(urlString: string): boolean {
-        try {
-            const parsedUrl = new URL(urlString);
-            return ALLOWED_DOMAINS.includes(parsedUrl.hostname);
-        } catch (e) {
-            return false;
-        }
-    }
-
     async fetchModel(uri: string): Promise<string> {
         if (!uri.startsWith('http://') && !uri.startsWith('https://')) {
             throw new Error(`Invalid URI scheme: ${uri}`);
         }
 
-        if (!this.isAllowedUrl(uri)) {
-            throw new Error(`SSRF Prevention: Domain not in allowlist. Received: ${uri}`);
+        // 1. Create a brand new URL object to break the CodeQL taint chain
+        let safeUrl: URL;
+        try {
+            safeUrl = new URL(uri);
+        } catch (e) {
+            throw new Error(`Malformed URL provided.`);
+        }
+
+        // 2. Validate the host of the new object
+        if (!ALLOWED_DOMAINS.includes(safeUrl.hostname)) {
+            throw new Error(`SSRF Prevention: Domain not in allowlist. Received: ${safeUrl.hostname}`);
         }
 
         const headers: Record<string, string> = {};
-
         if (process.env.EXTERNAL_TEMPLATE_TOKEN) {
             headers['Authorization'] = `Bearer ${process.env.EXTERNAL_TEMPLATE_TOKEN}`;
         }
@@ -39,13 +38,14 @@ export class HttpModelRetriever implements IModelRetriever {
         const timeoutId = setTimeout(() => controller.abort(), 5000);
 
         try {
-            const response = await fetch(uri, { 
+            // 3. Pass the validated URL object string, NOT the user's original 'uri'
+            const response = await fetch(safeUrl.toString(), { 
                 headers,
                 signal: controller.signal
             });
             
             if (!response.ok) {
-                throw new Error(`Failed to fetch model from ${uri}: ${response.statusText} (${response.status})`);
+                throw new Error(`Failed to fetch model from ${safeUrl.toString()}: ${response.statusText} (${response.status})`);
             }
 
             return await response.text();
--- a/server/handlers/retrievers/HttpModelRetriever.ts
+++ b/server/handlers/retrievers/HttpModelRetriever.ts
@@ -1,12 +1,39 @@
 import { IModelRetriever } from './IModelRetriever';
 
-const ALLOWED_DOMAINS = [
-    'models.accordproject.org',
-    'templates.accordproject.org',
-    'raw.githubusercontent.com'
-];
+export const MAX_FILE_SIZE = 1024 * 1024; 
 
-const MAX_FILE_SIZE = 1024 * 1024; 
+export function assertAllowedUrl(uri: string): URL {
+    if (!uri.startsWith('https://')) {
+        throw new Error(`Invalid URI scheme. Only https is allowed.`);
+    }
+
+    let safeUrl: URL;
+    try {
+        safeUrl = new URL(uri);
+    } catch (e) {
+        throw new Error(`Malformed URL provided.`);
+    }
+
+    if (safeUrl.port !== '' && safeUrl.port !== '443') {
+        throw new Error(`SSRF Prevention: Custom ports are not allowed.`);
+    }
+
+    const host = safeUrl.hostname;
+
+    if (
+        host !== 'models.accordproject.org' &&
+        host !== 'templates.accordproject.org' &&
+        host !== 'raw.githubusercontent.com'
+    ) {
+        throw new Error(`SSRF Prevention: Domain not in allowlist.`);
+    }
+
+    if (host === 'raw.githubusercontent.com' && !safeUrl.pathname.startsWith('/accordproject/')) {
+        throw new Error(`SSRF Prevention: Only official accordproject GitHub repositories are allowed.`);
+    }
+
+    return safeUrl;
+}
 
 export class HttpModelRetriever implements IModelRetriever {
     public getURISchemes(): string[] {
@@ -14,28 +41,7 @@ export class HttpModelRetriever implements IModelRetriever {
     }
 
     async fetchModel(uri: string): Promise<string> {
-        if (!uri.startsWith('https://')) {
-            throw new Error(`Invalid URI scheme. Only https is allowed.`);
-        }
-
-        let safeUrl: URL;
-        try {
-            safeUrl = new URL(uri);
-        } catch (e) {
-            throw new Error(`Malformed URL provided.`);
-        }
-
-        if (safeUrl.port !== '' && safeUrl.port !== '443') {
-            throw new Error(`SSRF Prevention: Custom ports are not allowed.`);
-        }
-
-        if (!ALLOWED_DOMAINS.includes(safeUrl.hostname)) {
-            throw new Error(`SSRF Prevention: Domain not in allowlist.`);
-        }
-
-        if (safeUrl.hostname === 'raw.githubusercontent.com' && !safeUrl.pathname.startsWith('/accordproject/')) {
-            throw new Error(`SSRF Prevention: Only official accordproject GitHub repositories are allowed.`);
-        }
+        const safeUrl = assertAllowedUrl(uri);
 
         const headers: Record<string, string> = {};
         if (process.env.EXTERNAL_TEMPLATE_TOKEN) {
@@ -44,12 +50,12 @@ export class HttpModelRetriever implements IModelRetriever {
 
         const controller = new AbortController();
         const timeoutId = setTimeout(() => controller.abort(), 5000);
-
+        
         try {
             const response = await fetch(safeUrl.toString(), { 
                 headers,
                 signal: controller.signal,
-                redirect: 'error' 
+                redirect: 'error'
             });
             
             if (!response.ok) {
@@ -62,7 +68,7 @@ export class HttpModelRetriever implements IModelRetriever {
             }
 
             const text = await response.text();
-
+            
             if (Buffer.byteLength(text, 'utf8') > MAX_FILE_SIZE) {
                 throw new Error('Model file exceeds the 1MB size limit after download.');
             }
--- a/server/handlers/retrievers/HttpModelRetriever.ts
+++ b/server/handlers/retrievers/HttpModelRetriever.ts
@@ -6,17 +6,18 @@ const ALLOWED_DOMAINS = [
     'raw.githubusercontent.com'
 ];
 
+const MAX_FILE_SIZE = 1024 * 1024; 
+
 export class HttpModelRetriever implements IModelRetriever {
     public getURISchemes(): string[] {
-        return ['http', 'https'];
+        return ['https'];
     }
 
     async fetchModel(uri: string): Promise<string> {
-        if (!uri.startsWith('http://') && !uri.startsWith('https://')) {
-            throw new Error(`Invalid URI scheme: ${uri}`);
+        if (!uri.startsWith('https://')) {
+            throw new Error(`Invalid URI scheme. Only https is allowed.`);
         }
 
-        // 1. Create a brand new URL object to break the CodeQL taint chain
         let safeUrl: URL;
         try {
             safeUrl = new URL(uri);
@@ -24,9 +25,16 @@ export class HttpModelRetriever implements IModelRetriever {
             throw new Error(`Malformed URL provided.`);
         }
 
-        // 2. Validate the host of the new object
+        if (safeUrl.port !== '' && safeUrl.port !== '443') {
+            throw new Error(`SSRF Prevention: Custom ports are not allowed.`);
+        }
+
         if (!ALLOWED_DOMAINS.includes(safeUrl.hostname)) {
-            throw new Error(`SSRF Prevention: Domain not in allowlist. Received: ${safeUrl.hostname}`);
+            throw new Error(`SSRF Prevention: Domain not in allowlist.`);
+        }
+
+        if (safeUrl.hostname === 'raw.githubusercontent.com' && !safeUrl.pathname.startsWith('/accordproject/')) {
+            throw new Error(`SSRF Prevention: Only official accordproject GitHub repositories are allowed.`);
         }
 
         const headers: Record<string, string> = {};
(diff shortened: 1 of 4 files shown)


rockaxorb13:
Hi @mttrbrts  and @dselman,

CodeQL is still flagging a Critical SSRF vulnerability on the fetch calls in both HttpModelRetriever.ts and HttpTemplateRetriever.ts. 

I have already implemented a strict URL parsing check and an explicit domain allowlist (models.accordproject.org, templates.accordproject.org, raw.githubusercontent.com). The code actively throws an error for any network requests outside of these domains, neutralizing the SSRF vector. 

Because CodeQL's taint-tracking isn't recognizing the allowlist validation as a sanitizer, this appears to be a strict false positive. Please let me know how to proceed with this one, can we dismiss this alert or is there any other way you want me to proceed?

niallroche:
continuing to look into this one, more detailed security checks may be necessary

rockaxorb13:
> continuing to look into this one, more detailed security checks may be necessary

Understood! Please tell if there's anything else that I could do

JayDS22 (commented):
Nice work adding the allowlist and URL-reconstruction pattern. That's a real step up from `HttpTemplateRetriever` (which has no allowlist at all). A few gaps I noticed while reading through, roughly ordered by SSRF blast radius. The CodeQL "1 critical alert" is almost certainly one of these.

### Critical

**1. `fetch()` follows redirects by default** (`server/handlers/retrievers/HttpModelRetriever.ts:44`)

The allowlist check runs against `safeUrl.hostname`, but `fetch(safeUrl.toString(), ...)` follows 3xx responses to any host. An allowlisted host (or a controlled repo under `raw.githubusercontent.com`) can respond `302` to `http://169.254.169.254/latest/meta-data/` and the request goes there. This is the most common SSRF bypass CodeQL flags on this shape.

Fix: pass `redirect: 'manual'` and either reject 3xx outright or re-validate each hop's hostname against the allowlist.

**2. Second-order SSRF via `ModelManager.updateExternalModels()`** (`server/handlers/sharedmodels.ts:17`)

The route calls `modelManager.updateExternalModels()` after `addCTOModel(ctoText, 'external.cto', true)`. Concerto's default `HttpModelFileLoader` fetches every model the CTO `import`s from arbitrary URLs with no allowlist. Attacker uploads a CTO that imports `http://internal-service/`, the allowlisted fetch of the parent succeeds, then Concerto downloads the internal URL on the same request. The direct-fetch allowlist doesn't cover this second hop.

`concerto-core`'s `basemodelmanager.js:450` shows `updateExternalModels(options, fileDownloader)` accepts a custom `fileDownloader`. Fix: build a downloader that reuses the same allowlist + redirect rules, and pass it in.

### High

**3. `raw.githubusercontent.com` is effectively an open allowlist** (`server/handlers/retrievers/HttpModelRetriever.ts:5`)

Any GitHub user can put arbitrary content at `raw.githubusercontent.com/<their-user>/<repo>/...`. That's not a bounded set. If the intent is Accord-Project-authored content, tighten to a path prefix check (e.g. only accept `raw.githubusercontent.com/accordproject/`, or a known-orgs list).

**4. No response size cap** (`server/handlers/retrievers/HttpModelRetriever.ts:48`)

`await response.text()` reads the full body into memory. Even from an allowlisted host, an attacker with write access to any allowlisted GitHub repo can serve a huge file. Add a `Content-Length` check up front and a streaming read with a hard cap (say 1 MB for a `.cto`).

### Medium

**5. `http://` allowed alongside `https://`** (`server/handlers/retrievers/HttpModelRetriever.ts:9, 15`)

All allowlisted domains serve HTTPS. Dropping `http://` from `getURISchemes()` closes the door on MITM downgrade and any internal HTTP-only service that happens to bind on one of the allowlisted hostnames' unusual ports.

**6. DNS rebinding is unmitigated** (`server/handlers/retrievers/HttpModelRetriever.ts:20, 44`)

The hostname check happens at parse time; DNS resolves later during `fetch()`. A controlled DNS server can return the allowlisted hostname's real IP first (pass allowlist), then an internal IP on the actual connect. Mitigation is resolve DNS explicitly, reject private/loopback IPs, fetch by IP with the Host header set. Fiddly to get right. If it's out of scope for this PR, worth a follow-up issue.

### Low

**7. Error message leak in caller** (`server/handlers/sharedmodels.ts:29-30`)

`{ details: error.message }` can leak URL, resolved hostname, or internal fetch state. Map known cases to safe strings; log the raw error server-side only.

**8. `HttpTemplateRetriever.ts` has none of these protections**

Out of scope for this PR, but for maintainer awareness: the existing `HttpTemplateRetriever` (used by `POST /agreements`) does `fetch(uri, { headers })` with no allowlist, no redirect handling, no size cap. Whatever pattern lands here should be ported over.

---

None of these block the direction. The allowlist + URL-reconstruction approach is the right shape; (1) and (2) are what actually close the SSRF surface.

rockaxorb13:
Thanks alot for the detailed reply! I am extremely sick at the moment but I'll start working and join meeting regularly ASAP!!

JayDS22:
No rush, take care of your health first. When you're back, the two critical items (redirect-following at `server/handlers/retrievers/HttpModelRetriever.ts:44` and second-order SSRF via `updateExternalModels`) are the ones that actually close the CodeQL alert; the rest is defense-in-depth.

niallroche:
Thanks for continuing this @rockaxorb13, and sorry for the slow turnaround. The shape of the feature is right — `IModelRetriever` separated from `ITemplateRetriever` makes sense given `.cto` text vs. `.cta` buffers, and validating through `ModelManager` before insert is the correct instinct. The domain allowlist, scheme check, and 5s abort timeout are all good additions since the first pass.

I've gone through the whole path (handler → retriever → parse) rather than just the retriever, and there's one issue that I think needs a design decision before the code is finishable, plus a few concrete fixes.

**1. `updateExternalModels()` re-opens SSRF past the allowlist**

In `sharedmodels.ts`:

```ts
const modelFile = modelManager.addCTOModel(ctoText, 'external.cto', true);
await modelManager.updateExternalModels();
```

The allowlist in `HttpModelRetriever` guards the entry fetch — the .cto the caller names. But a .cto can declare arbitrary imports, and `updateExternalModels()` resolves them through Concerto's own `FileDownloader`, which never sees `ALLOWED_DOMAINS`:
```
import org.example.Thing from http://169.254.169.254/latest/meta-data/…
```

So a model served from an allowlisted host can pull second-order fetches to anywhere. No redirect needed — transitive import resolution is a designed feature of Concerto, and it bypasses the allowlist by construction. This is the one I'd treat as blocking, because it isn't a missing flag; it means the guard is on the wrong layer.

Two ways out, both bigger than a patch:

- Parse the model's imports first and reject any whose URL isn't allowlisted, before calling `updateExternalModels()`; or
- Inject an allowlist-aware loader/FileDownloader into the ModelManager so every fetch it makes goes through the same check.
Worth deciding which before more code goes in.

Redirects aren't constrained

fetch() follows redirects by default and only the initial hostname is validated, so an allowlisted host can 302 to an internal address:
```
const response = await fetch(safeUrl.toString(), {
    headers,
    signal: controller.signal,
    redirect: 'manual',   // or 'error'
});
```

With 'error' it rejects on any redirect; with 'manual' you can re-validate Location against the allowlist before following. Related: `EXTERNAL_TEMPLATE_TOKEN` is attached based on the pre-redirect host, so an unconstrained redirect can forward the bearer token to the redirect target.

- No response-size cap. response.text() reads unbounded; updateExternalModels() amplifies this since one model can pull many files. A Content-Length check or capped read would help.
- http:// is accepted alongside https. For fetching trusted models, https-only avoids plaintext/MITM.
- Port isn't validated. URL.hostname excludes the port, so `https://models.accordproject.org:22/`… passes the allowlist check.
- Hostname allowlists are checked pre-DNS, so they don't stop rebinding or a host that resolves to a private address. Not necessarily in scope here, but worth knowing the allowlist alone isn't a complete control.

Also currently `POST /sharedmodels` isn't behind auth in the reference implementation,

rockaxorb13:
Understood, I will be discharged tomorrow and will start working ASAP

rockaxorb13:
Hey @JayDS22, @niallroche! I am back from the hospital and ready to rejoin! Thank you so much for you patience! Please let me know if the changes I have made are good to go or not. I see that the CodeQL tests are still failing, though.

JayDS22:
Welcome back @rockaxorb13, hope you're feeling better.

On the CodeQL findings at `server/handlers/retrievers/HttpModelRetriever.ts:19, 45, 53`: those are false positives. The retriever has a real SSRF gate that runs before the fetch:

- `ALLOWED_DOMAINS` allowlist (L3-7)
- Port 443 only (rejects custom ports)
- `accordproject/` path prefix check for `raw.githubusercontent.com`
- `https` scheme restriction

CodeQL's static SSRF query cannot see a runtime allowlist as a sanitizer, so it fires on any user-input-to-`fetch` flow regardless of the guards.

Two ways to resolve:

**Option A: dismiss on the GitHub Security tab** as false positive, with a justification note pointing at the `ALLOWED_DOMAINS` allowlist. Repo maintainers (@niallroche / @dselman) can do this from the Security > Code scanning alerts view. Fastest.

**Option B: refactor into a single `fetchAllowlistedUrl(uri)` helper** that inlines the validate-then-fetch flow. CodeQL sees the sanitization on the same call path and stops firing. More code but no ongoing dismissal maintenance. Would look something like:

```typescript
async function fetchAllowlistedUrl(uri: string): Promise<Response> {
    const safeUrl = validateAndParse(uri); // throws on any allowlist violation
    return fetch(safeUrl);
}
```

I'd lean Option A here since the current guards are already correct and readable. Once CodeQL is off the blocker, happy to do a full review pass on the delta since my Jul 24 comments.

niallroche:
Welcome back @rockaxorb13 — no worries at all about the timing, health comes first. And thank you for taking the whole list seriously; this is a big step up. Almost everything from the last pass is genuinely closed:

- Redirects: `redirect: 'error'` is exactly right, and as a bonus it also closes the bearer-token-forwarding case, since no redirect can happen now.
- `raw.githubusercontent.com` narrowed to `/accordproject/` paths — good.
- https-only, port-443-only, and the 1 MB cap (Content-Length *and* post-read byte length) all look correct.

On the one that mattered most — the transitive-import SSRF — you've got the right idea (inject a downloader so import resolution goes through the allowlist too), but the wiring doesn't actually take effect, and I think it's also what's keeping CI red. `SecureFileDownloader` only defines `getFile()`:

​```ts
class SecureFileDownloader {
    async getFile(url: string): Promise<string> { ... }
}
await modelManager.updateExternalModels({}, secureDownloader as any);
​```

But `updateExternalModels(options, fileDownloader)` never calls `getFile`. Internally it does:

​```js
const externalModels = await fileDownloader.downloadExternalDependencies(this.getModelFiles(), options);
​```

So Concerto calls `downloadExternalDependencies(...)`, which your object doesn't have — the `as any` is the only reason it compiles. At runtime that means the model's imports are resolved by a downloader that either throws (`downloadExternalDependencies is not a function`) or, if the cast slips past, never consults your allowlist at all. Either way the allowlist isn't enforced on imports yet.

The interface you want to implement is `downloadExternalDependencies`, not `getFile`. The idiomatic route is Concerto's own `ModelFileDownloader` wrapping a custom `FileLoader` — the loader exposes `accepts(url)` / `load(url, options)`, and you put the allowlist/scheme/port/path checks (the same ones already in `HttpModelRetriever`) inside `accepts`/`load` so every transitive fetch goes through them. Then pass `new ModelFileDownloader(secureLoader)` as the second arg and drop the `as any`. Factoring the validation out of `HttpModelRetriever` into a shared `assertAllowed(url)` that both the entry fetch and the loader call would keep the two paths honest.

The thing that'll prove it's actually wired: a test where a `.cto` served from an allowlisted host declares `import … from https://<not-allowlisted-host>/x.cto` and the request is rejected. That test failing today, and passing after, is the whole ballgame for this PR.

Happy to look again as soon as that's in as I think we are getting closer.

rockaxorb13:
Hi @JayDS22, @niallroche! I am recovering well! Hope to catch you in the next wg call.

Thanks a lot for the detailed feedback and the clear explanation of how Concerto expects transitive dependencies to be resolved!

I implemented the custom secure downloader using downloadExternalDependencies alongside a SecureFileLoader. Both entry fetches and transitive imports now pass through the shared assertAllowedUrl validation in HttpModelRetriever, ensuring all domain allowlists, HTTPS/port rules, path constraints, and size restrictions are strictly enforced across all steps.

Also, for the automated Unit Test, I created server/handlers/sharedmodels.test.ts to test the exact scenario you described. The test mocks a request for a model served from an allowlisted host (models.accordproject.org) that contains a transitive import statement pointing to a non-allowlisted domain. It asserts that the request is rejected with a 400 status code and confirms that the non-allowlisted host is never fetched.


please direct me on how to proceed further!

JayDS22:
Nice, this is the right shape and hits the `downloadExternalDependencies` path Niall named as the real risk. CodeQL is still failing at `HttpModelRetriever.ts:55` on `fetch(safeUrl.toString(), ...)` because the taint tracker doesn't recognize `assertAllowedUrl` as a sanitizer. Two ways to close:

**A. Suppression + justification** (fastest):

```ts
// codeql[js/request-forgery] Host validated against ALLOWED_HOSTS in assertAllowedUrl above.
const res = await fetch(safeUrl.toString(), { ... });
```

**B. Refactor so the taint tracker sees the sanitization:**

```ts
assertAllowedUrl(url);
// Reconstruct URL from allowlisted host + validated path so CodeQL sees a literal-shaped URL.
const finalUrl = `https://${url.host}${url.pathname}${url.search}`;
const res = await fetch(finalUrl, { ... });
```

I'd go with **B**. Cleaner long-term, no suppression comment to maintain across CodeQL bumps, and the reconstruction is defensive since it drops any URL bits `assertAllowedUrl` didn't validate.

Minor nits:
- The test asserts `fetch.toHaveBeenCalledTimes(1)` but `SecureFileLoader.accepts` short-circuits before `fetch` in the evil path, so the assertion is technically correct but doesn't prove `getImports()` actually surfaced the malicious import. Consider adding one assertion on the parsed CTO to confirm the import reached `updateExternalModels`.
- Missing trailing newlines on the 3 new files.

rockaxorb13:
@JayDS22 @niallroche, I made the changes but I see that the Code scanning results are still failing, please let me know how to proceed! Thanks!

JayDS22:
@rockaxorb13 the guard itself is correct: scheme + port + host allowlist + accordproject-only path prefix for `raw.githubusercontent.com`. CodeQL is failing not because the fix is wrong, but because CodeQL's dataflow can't prove that `assertAllowedUrl` sanitizes the URL by the time it reaches `fetch(safeUrl.toString(), ...)` at line 55. It's a well-known CodeQL false positive on custom-validator patterns.

Two real options:

**Option 1 (fastest, needs maintainer):** dismiss the alert in the Security tab with reason 'False positive - URL is validated by `assertAllowedUrl` (scheme + port + host allowlist)'. Only @niallroche or another maintainer with `security-events: write` can do this. Once dismissed, the alert won't re-fire on the same rule + location.

**Option 2 (goes green without maintainer action):** restructure so CodeQL sees a hardcoded prefix. Something like:

```ts
const ALLOWED_HOSTS = ['models.accordproject.org', 'templates.accordproject.org', 'raw.githubusercontent.com'] as const;
type AllowedHost = typeof ALLOWED_HOSTS[number];

// Return the discriminated host, not just the URL, so the fetch site can rebuild from a literal:
export function assertAllowedUrl(uri: string): { host: AllowedHost; pathname: string; search: string } { ... }

// Then in fetchModel:
const { host, pathname, search } = assertAllowedUrl(uri);
const response = await fetch(`https://${host}${pathname}${search}`, { ... });
```

CodeQL treats the template literal as a fresh flow with a hardcoded scheme + allowlist-typed host, and the taint gets scrubbed. Slightly uglier but no maintainer needed.

I'd go option 1 if @niallroche is OK dismissing. Option 2 as fallback.

Not blocking on my end. The security posture is already correct.

rockaxorb13:
@JayDS22 Understood! Thank you very much for the quick response! :D

I wanted to ask if there's any other work I could pick up or be assigned here or in other repositories?

niallroche:
Welcome back @rockaxorb13, I hope all is well. The hardening in the last push addresses issues such as: `redirect: 'error'`, https-only, port-443-only, the 1 MB cap, and narrowing `raw.githubusercontent.com` to `/accordproject/`. 

Two things that I can see, one old and one new:

**1. The transitive-import fix still isn't wired (unchanged).** `SecureFileDownloader` only defines `getFile()`, but `ModelManager.updateExternalModels(options, fileDownloader)` never calls `getFile` — it calls `fileDownloader.downloadExternalDependencies(this.getModelFiles(), options)` (verified against concerto source). The `as any` is the only reason it compiles; at runtime the imports resolve through a downloader that either throws or bypasses the allowlist. Implement `downloadExternalDependencies` — the idiomatic route is a custom `FileLoader` (allowlist/scheme/port/path checks in `accepts`/`load`) wrapped in concerto's `ModelFileDownloader`, passed as the second arg. The proof it's wired: a test where an allowlisted-host `.cto` declares `import … from https://<non-allowlisted-host>/x.cto` and the request is rejected.

**2. New: `main` moved to concerto-core/util 4.x (#236 merged).** This branch was written against 3.x, so it'll need a rebase onto 4.x — and note `ModelManager` construction changed (the `strict: true` option is gone in 4.x, since versioned namespaces are now mandatory). Worth verifying the downloader interface against concerto-util 4.x's `ModelFileDownloader` while you're rebasing, since that's the version that'll actually run.

Happy to re-review as soon as (1) is wired against the 4.x baseline.

rockaxorb13:
@niallroche Understood! I shall get to this as soon as possible.

I am doing much better now. Thanks for asking :)

rockaxorb13:
Hi @niallroche, I have an update and a doubt in my work so far: 

**_Update:_**
I've successfully rebased my branch onto the latest `main` to pick up the Concerto v4 upgrades (#236), and I've removed the `strict: true` flag from the `ModelManager` constructor as requested.

**_Doubt:_**
Regarding the transitive-import fix (Issue 1): I completely agree with the approach and want to do this the idiomatic way by wrapping our custom loader in Concerto's native downloader. However, I've hit a small roadblock with the v4 API. 

After syncing and inspecting the local `node_modules`, I'm having trouble locating `ModelFileDownloader` in the exports of `@accordproject/concerto-util` (or `concerto-core`). I do see `FileDownloader` in `concerto-util`, but its constructor expects two arguments: `(fileLoader: any, getExternalImports: any)`. 

Could you point me to where `ModelFileDownloader` is exported in the 4.x baseline, or let me know if I should be using `FileDownloader` directly (and how best to pass the `getExternalImports` function)? 

Happy to push the finalized fix as soon as I can get that wired up correctly, or we can chat about it in the next WG call if that's easier!

rockaxorb13:
commenting to remove the stale status

niallroche:
Welcome back @rockaxorb13, and as discussed on the working group call last week, thanks for continuing to push this forward. I re-reviewed the current head and also did a merge rehearsal onto current `main`.

The good news: the main security shape is much better now. `redirect: 'error'`, https-only, port-443-only, the 1 MB size cap, and `raw.githubusercontent.com` narrowed to `/accordproject/` are all the right direction. I also verified that the new focused sharedmodels test passes, and the merged tree builds against current `main`.

A few things still need tightening before merge:

1. **Use Concerto’s recursive downloader rather than the hand-rolled one**

`SecureDownloader.downloadExternalDependencies()` now exists, which addresses the previous “not wired” blocker, but it only scans imports from the initial model files. Concerto’s own `FileDownloader` recursively walks imports from downloaded files too. With the current version, a valid chain like `A.cto -> B.cto -> C.cto` will fetch `B` but not `C`, and then fail validation.

I’d suggest using `@accordproject/concerto-util.FileDownloader` with your `SecureFileLoader`, passing a `getExternalImports` function, rather than reimplementing the recursion.

2. **CodeQL is still blocking CI**

The runtime URL guard looks sensible, but CodeQL still flags `fetch(safeUrl.toString(), ...)`. I now think Jay’s Option 2 is the better route here: have `assertAllowedUrl()` return a validated `{ host, pathname, search }` shape where `host` is an allowlisted type, then reconstruct the fetch URL from that:

```ts
const response = await fetch(`https://${host}${pathname}${search}`, { ... });
```

That should make the taint flow clear to CodeQL and avoid needing a maintainer dismissal.

3. **Add a success-path test**

The current test covers the malicious transitive import rejection, which is good. Please also add a positive test showing a valid external CTO fetch is parsed and rewritten into the expected `org.accordproject.protocol@1.0.0.CtoModel` payload before CRUD insertion. That route mutates `req.body`, so it’s worth pinning the happy path too.

4. **Small cleanup**

`git diff --check` still reports trailing whitespace and missing final newlines in the new/changed files. Please clean those up before the next push.
Once those are addressed I think this will be in much better merge shape.

rockaxorb13:
@niallroche done! Please let me know how to proceed further, there seems to be a failure for codeQL but everything else looks good!

niallroche:
one point to note @rockaxorb13 is that a URL like `https://evil.com@models.accordproject.org/x.cto` passes both the `startsWith('https://')` check and the `hostname` allowlist check (`hostname` returns `models.accordproject.org`). `fetch` would send the request to the allowlisted host, not to `evil.com`, so it's not a true SSRF — but unexpected `userinfo` could leak in headers. Worth a one-line fix: add something like 
`if (safeUrl.username || safeUrl.password) throw new Error(...)` inside `assertAllowedUrl`.

Also consider adding the username/password check to assertAllowedUrl

rockaxorb13:
@niallroche done! Thanks for pointing that out! Please let me know if my implementation is acceptable or not!

PR #177, opened 2026-05-27
fix(crud): harden pagination query parsing

Rishabh060105:
## Summary

Follow-up to #169 for the pagination sanitization fix related to #168.

This PR hardens pagination query parsing in `buildCrudRouter` so malformed `page` and `limit` query params cannot leak unintended values into SQL pagination.

## References

- follows up on #169
- related to #168

## What Changed

- accept only whole decimal strings for `page` and `limit`
- fall back to safe defaults when pagination params are invalid
- normalize repeated query params by using the first value
- preserve the existing `limit <= 100` clamp
- add focused unit coverage for invalid, partially numeric, and repeated pagination params

## Root Cause

The original fix in #169 handled the `NaN` case from `parseInt`, but it still allowed partially numeric values like `12abc` and alternate numeric forms like `0x10` to parse into unintended pagination values.

## Impact

This makes CRUD pagination behavior more predictable and prevents malformed query strings from producing surprising offsets or limits.

## Validation

- `npm run build` in `server`
- `npm test` in `server`
- `npm test` in repository root

PR #208, opened 2026-07-11
fix(crud): reject non-strict-numeric :id up-front (closes #162)

### LINKED ISSUE #162: bug: Numeric `:id ` routes vulnerable to partial parses.
# Bug Report 🐛

## Expected Behavior

the api should reject invalid, non-numeric `:id` parameters (like `1abc`) with something like 404 not found or 400. valid requests with strict numeric id should process normally.

## Current Behavior

the code uses `parseInt(req.params.id)` to parse the id from the route. `parseInt` allows partial parses which can cause something like: `1abc`-----> parsed to `1`. so some route like `http://localhost:9000/templates/1abc` will resolve to `http://localhost:9000/templates/1`. which is unexpected behaviour.

Image

## Steps to Reproduce

1. start the server.
2. create a resource (e.g., an agreement) so that an item with ID `1` exists in the database.
3. send a GET request to the endpoint with an invalid id ( `GET /agreements/1abc`).
4. observe that the server successfully processes the request for ID `1` instead of returning an error for `1abc`.

JayDS22:
## Summary

Closes #162. `parseInt(req.params.id)` returns `1` for the string `'1abc'`, so before this guard:

- `PUT /agreements/1abc` with a valid body silently updated row `1`
- `DELETE /templates/1abc` silently deleted row `1`
- `GET /templates/1abc` silently returned row `1`

## What this PR does

Adds a route-level `router.param('id', ...)` middleware in `buildCrudRouter` that rejects anything that is not a strict decimal integer with `404` on integer-keyed tables. UUID-keyed tables keep the existing pass-through so the DB layer can surface malformed UUIDs.

The guard runs before any body validation, DB read, DB mutation, or side-effectful trigger, so a bad `:id` never touches the DB and never blocks on Concerto validation of the body.

## Tests

7 new cases appended to `crud.test.ts` covering GET, PUT, DELETE with:

- `1abc` (parseInt partial parse, the case from the issue)
- `1.5` (float)
- `-1` (negative)
- `0x10` (hex)
- `1e2` (scientific)

PUT and DELETE cases also assert that `dbMock.update` / `dbMock.delete` were never called, pinning the safety property that the guard runs before any DB mutation. The accepted-path regression is already covered by the existing `DELETE /:id returns 200 when resource exists` case.

## Also in this PR

Fixes a pre-existing test bug in `crud.test.ts` that imported `DbTemplate` from `../db/schema`. `DbTemplate` is a local alias in `agreements.ts` (`import { Template as DbTemplate }`), not a named export from the schema module, so the two tests that used it got `table: undefined`. They silently passed because the exercised code path never touched `table` before returning. The new `:id` middleware accesses `table.id.columnType` up-front, so the real export name is required now. This is a one-word fix (`DbTemplate` -> `Template`) with an inline comment explaining why.

## Validation

- `npm test` in `server/`: **98/98 passing** (16 in `crud.test.ts`)
- `npx tsc --noEmit` in `server/`: clean

## Non-goals

- UUID format validation is out of scope; the guard skips UUID tables and leaves DB-layer errors in place
- No changes to `parseInt` call sites inside handlers; the middleware guarantees strict-decimal input reaches them
- No overlap with #187 (`defaultWhereClause` SQL injection guard) or #177 (`parseQueryParams` hardening) — different functions in the same file

## Author Checklist

- [x] DCO sign-off provided
- [x] Tests added and passing
- [x] Typecheck clean

### DIFF
--- a/server/handlers/crud.test.ts
+++ b/server/handlers/crud.test.ts
@@ -143,16 +143,18 @@ describe('PUT /:id validation', () => {
             error: { errors: [{ message: 'Custom validation failed' }] }
         });
 
-        // `DbTemplate` is a local alias in agreements.ts (import { Template as DbTemplate }),
-        // not a named export from db/schema. This block previously passed with `table:
-        // undefined` because the custom-validation short-circuit returned before any code
-        // touched `table`. The :id middleware added for #162 accesses `table.id.columnType`
-        // up-front, so the real export name is required now.
+        // `DbTemplate` is not a named export from db/schema. It's a local alias in
+        // agreements.ts (`import { Template as DbTemplate }`). Re-aliasing here keeps
+        // the reader-facing name consistent with agreements.ts while pulling the real
+        // export. The previous direct `DbTemplate` import resolved to `undefined` and
+        // the tests only passed because the custom-validation short-circuit returned
+        // before any code touched `table`. The :id middleware added for #162 accesses
+        // `table.id.columnType` up-front, so a real schema object is required now.
         const { buildCrudRouter } = require('./crud');
-        const { Template } = require('../db/schema');
+        const { Template: DbTemplate } = require('../db/schema');
 
         const testRouter = buildCrudRouter({
-            table: Template,
+            table: DbTemplate,
             validateBody: {
                 // no schema — only custom
                 custom: (body: any) => valModule.concertoValidation('Template', body)
@@ -337,11 +339,11 @@ describe('POST without validateBody', () => {
         });
 
         const { buildCrudRouter } = require('./crud');
-        const { Template } = require('../db/schema');
+        const { Template: DbTemplate } = require('../db/schema');
 
         // No validateBody at all
         const testRouter = buildCrudRouter({
-            table: Template,
+            table: DbTemplate,
             typeName: 'templates',
         });
 
--- a/server/handlers/crud.test.ts
+++ b/server/handlers/crud.test.ts
@@ -143,12 +143,16 @@ describe('PUT /:id validation', () => {
             error: { errors: [{ message: 'Custom validation failed' }] }
         });
 
-        // Import buildCrudRouter and DbTemplate directly
+        // `DbTemplate` is a local alias in agreements.ts (import { Template as DbTemplate }),
+        // not a named export from db/schema. This block previously passed with `table:
+        // undefined` because the custom-validation short-circuit returned before any code
+        // touched `table`. The :id middleware added for #162 accesses `table.id.columnType`
+        // up-front, so the real export name is required now.
         const { buildCrudRouter } = require('./crud');
-        const { DbTemplate } = require('../db/schema');
+        const { Template } = require('../db/schema');
 
         const testRouter = buildCrudRouter({
-            table: DbTemplate,
+            table: Template,
             validateBody: {
                 // no schema — only custom
                 custom: (body: any) => valModule.concertoValidation('Template', body)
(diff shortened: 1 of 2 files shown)


server/handlers/crud.test.ts
```diff
@@ -143,12 +143,16 @@ describe('PUT /:id validation', () => {
...
 
-        // Import buildCrudRouter and DbTemplate directly
+        // `DbTemplate` is a local alias in agreements.ts (import { Template as DbTemplate }),
+        // not a named export from db/schema. This block previously passed with `table:
+        // undefined` because the custom-validation short-circuit returned before any code
+        // touched `table`. The :id middleware added for #162 accesses `table.id.columnType`
+        // up-front, so the real export name is required now.
         const { buildCrudRouter } = require('./crud');
-        const { DbTemplate } = require('../db/schema');
+        const { Template } = require('../db/schema');
```
apoorv7g: Why has this been changed from `DbTemplate` to `Template`? What purpose does it achieve ?

apoorv7g (commented):
We should keep the old variable name because it makes it much clearer what the template is actually for

server/handlers/crud.test.ts
JayDS22: `DbTemplate` isn't a named export from `db/schema`. It's a local alias inside `agreements.ts` (`import { Template as DbTemplate }`). The two tests importing it were getting `undefined` silently and only passed because the exercised code path never touched `table` before returning. The new `:id` middleware accesses `table.id.columnType` up-front, so the actual export name is required now.

There's an inline comment on the import explaining this, plus the "Also in this PR" section of the description.

JayDS22:
Follow-up on the review comment about keeping the old name. Noted the readability concern. Alias resolution keeps `DbTemplate` at the call site while pointing at the real export: `const { Template: DbTemplate } = require('../db/schema')`. Can apply that if the local name is what matters. The line-comment thread on `crud.test.ts:152` has the technical detail on why the literal import can't stay.

server/handlers/crud.test.ts
apoorv7g: Alright, I overlooked it, also here is a better way we could handle the requirement.

`const { Template: DbTemplate } = require('../db/schema');`  

Let me know if this works out well.

apoorv7g:
Sorry, I didn't check this.

> Follow-up on the review comment about keeping the old name. Noted the readability concern. Alias resolution keeps `DbTemplate` at the call site while pointing at the real export: `const { Template: DbTemplate } = require('../db/schema')`. Can apply that if the local name is what matters. The line-comment thread on `crud.test.ts:152` has the technical detail on why the literal import can't stay.

 Yep that's what I mean.

JayDS22:
Applied. `19ffbff` aliases at the import site: `const { Template: DbTemplate } = require('../db/schema')` in both blocks, `table: DbTemplate` at the call site. Comment updated to explain the aliasing. Local test run stays 16/16 green.

apoorv7g (approved):

niallroche (approved):
Approved. Two notes for the record, since this touches resource identity:

This guard hardens how the reference implementation addresses resources — via the Postgres serial id on the /:id routes. That integer is an RI storage detail (the DB primary key), not part of the APAP protocol. The protocol identifies Template, Agreement, and SharedModel by uri (identified by uri in protocol.cto); the OpenAPI contract paths are /{uri}, typed string). While the RI uses Postgress, an APAP compliant server backed by a document store, content-addressed store, or other is free to use any string identity and remains conformant.

So this PR should be read as "the RI rejects malformed integer ids before they reach the DB," not "APAP resources are integer-identified." The longer-term alignment is likely to need some additional work in routing the RI on uri to match the published contract .

PR #211, opened 2026-07-18
feat(services): port templateService + rewire mcp.ts template loops (slice 1)

JayDS22:
## Summary

Slice 1 of the shared-service-layer port from [apap-mcp-poc](https://github.com/JayDS22/apap-mcp-poc). Adds the first service module upstream and rewires all four MCP template call sites in `mcp.ts` to call it directly instead of going through the internal `makeApiRequest(\`${API_BASE_URL}/templates...\`)` HTTP loop.

This is the smallest slice that both introduces the new pattern AND proves the payoff (four fewer internal HTTP loops in `mcp.ts`). Slice 2 will bring across the agreement service; slice 3 will rewire the REST `/templates` route to call the same functions.

### `server/db/client.ts` (new, 8 lines)
`Database` type alias for `PostgresJsDatabase<typeof schema>`. Services take this as their first argument so they stay transport-agnostic and both MCP + REST callers get the same typed handle.

### `server/services/templateService.ts` (new, 69 lines)
Six exported functions: `listTemplates`, `getTemplateById`, `getTemplateByUri`, `createTemplate`, `updateTemplate`, `deleteTemplate`. All take `db` as the first arg, return typed rows, and throw `TemplateNotFoundError` / `TemplateDuplicateError` from the `ServiceError` hierarchy introduced in #184. Zero imports from Express or the MCP SDK.

### `server/services/templateService.test.ts` (new, 211 lines)
13 unit tests over a fluent Drizzle query-builder mock. Covers success and not-found paths for all six functions plus the 23505 unique-violation and generic re-throw branches on `createTemplate`.

### `server/handlers/mcp.ts` — four MCP call sites rewired
- `getServer` signature changes from `()` to `(db: Database)`; both callers pass `res.locals.db`
- `getTemplates` module fn now uses `listTemplates(db)` instead of `makeApiRequest`
- `apap://templates/{templateId}` list callback uses `listTemplates(db)`
- `apap://templates/{templateId}` read callback uses `getTemplateById(db, ...)`
- `getTemplate` tool uses `getTemplateById(db, ...)`

Same strict-numeric guard as #208 (`/^\d+$/`) applied to the `templateId` path segment before calling `getTemplateById`, so a malformed resource URI resolves to `TemplateNotFoundError` instead of `NaN`. The existing `serviceErrorToResourceError` / `serviceErrorToCallToolResult` mappers convert the typed errors to protocol-appropriate responses.

## Not in this slice

- Agreement service port (slice 2, larger surface: trigger + convert + resolve)
- REST `/templates` route rewire via `buildCrudRouter` (slice 3, would collide with #192 middleware surface until it lands)
- `subscriptionRegistry` port (separate slice, POC-only for now)

## Validation

- `npm test` in `server/`: 9/9 suites pass, 121 tests total (13 new)
- `npx tsc --noEmit` clean
- REST `/templates` route untouched, so #187 SQL injection guard and #208 `:id` middleware are preserved

## Author Checklist

- [x] DCO sign-off provided
- [x] Tests added and passing
- [x] Typecheck clean
- [x] No transport-specific imports in `services/`
- [x] `ServiceError` subclasses used for typed errors (no bare `throw new Error(...)` from the service)

### DIFF
--- a/server/db/client.ts
+++ b/server/db/client.ts
@@ -1,0 +1,8 @@
+import type { PostgresJsDatabase } from 'drizzle-orm/postgres-js';
+import * as schema from './schema';
+
+// The RI wires the shared Drizzle handle onto `res.locals.db` at request time
+// (see server/index.ts). Services take this typed handle as their first arg so
+// they stay transport-agnostic — the MCP handler and REST routes can both call
+// the same service function without going through an internal HTTP loop.
+export type Database = PostgresJsDatabase<typeof schema>;
--- a/server/handlers/mcp.ts
+++ b/server/handlers/mcp.ts
@@ -16,6 +16,8 @@ import {
     AgreementTriggerError,
     UpstreamApiError,
 } from '../services/errors';
+import { listTemplates, getTemplateById } from '../services/templateService';
+import type { Database } from '../db/client';
 
 const HOST = process.env.HOST || 'localhost';
 const PORT = parseInt(process.env.PORT || '9000', 10);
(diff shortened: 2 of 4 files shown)


niallroche (changes requested):
hi Jay, a couple of observations listTemplates(db) does db.select().from(Template) with no limit. Framed purely on the reference impl: the apap://templates MCP resource previously went through parseQueryParams, which clamped to ≤100 rows; it now returns the entire table on every read. That's a token-budget and scalability regression for the resource path, and it means the per-resource ttlMs/cacheScope hints from #201 now annotate an unbounded payload. Recommend the primitive carry bounds:

```typescript
export async function listTemplates(
  db: Database,
  opts: { limit?: number; offset?: number } = {},
): Promise<TemplateRow[]> {
  const limit = Math.min(100, Math.max(1, opts.limit ?? 100));
  const offset = Math.max(0, opts.offset ?? 0);
  return db.select().from(Template).limit(limit).offset(offset);
}
```

Worth fixing here rather than in slice 3, so the REST unification inherits a bounded primitive instead of having to reintroduce paging in-memory over a full-table fetch.

2. Forward-looking, for slice 3: the service path intentionally skips the Express chain, which is fine for reads. When the REST write routes get unified onto createTemplate/updateTemplate, please make sure the zod + Concerto validation the CRUD router runs today is preserved in or ahead of the service — just flagging now so it doesn't fall out during the unification.

Everything else here looks good — the strict-numeric id guard carried over from #208 and the typed TemplateNotFoundError are both nice.

PR #213, opened 2026-07-18
feat(services): port agreementService CRUD half + rewire mcp.ts agreement lookups (slice 2)

JayDS22:
## Summary

Slice 2 of the shared-service-layer port from [apap-mcp-poc](https://github.com/JayDS22/apap-mcp-poc). Adds `agreementService` upstream and rewires the four MCP agreement lookup call sites in `mcp.ts` to call it directly instead of going through the internal `makeApiRequest(\`${API_BASE_URL}/agreements...\`)` HTTP loop.

## Stack

**This PR is stacked on top of #211 (slice 1).** Its base is `main` for reviewability, but its commit graph includes the slice 1 commit for the `Database` type and `getServer(db)` signature. When #211 merges, GitHub will automatically collapse this PR's diff to just the three agreement-related files below.

## Scope choice: CRUD half only

Only `listAgreements` and `getAgreementById` are ported from the POC's `agreementService.ts`. The POC's `convertAgreement` and `triggerAgreement` are simplified stubs that do not run the real `@accordproject/template-engine`. The RI's REST route at `server/handlers/agreements.ts` already runs `TemplateArchiveProcessor` for both convert and trigger, so porting the stubs would regress that.

Convert and trigger deserve their own slice (call it 2b) that wraps the real template engine as a service rather than replacing it with a POC stub.

### `server/services/agreementService.ts` (new, 25 lines)
Two exported functions: `listAgreements(db)` and `getAgreementById(db, id)`. Both take `db` as the first arg, return typed rows, and throw `AgreementNotFoundError` from the `ServiceError` hierarchy introduced in #184. Zero transport-specific imports.

### `server/services/agreementService.test.ts` (new, 85 lines)
4 unit tests using the same fluent Drizzle mock pattern established in `templateService.test.ts` (slice 1). Covers success and not-found paths for both functions.

### `server/handlers/mcp.ts` — four call sites rewired
- `getAgreement` module fn takes `db: Database`, uses `getAgreementById`
- `getAgreements` module fn takes `db: Database`, uses `listAgreements`
- `apap://agreements/{agreementId}` list callback uses `listAgreements(db)`
- `getAgreement` tool uses `getAgreementById(db, ...)`

Same strict-numeric guard as #208 (`/^\d+$/`) applied to the `agreementId` path segment before calling the service, so a malformed resource URI resolves to `AgreementNotFoundError` instead of `NaN`. The #128 fix (agreement's own `uri` field must not overwrite the MCP resource URI) is preserved.

## Not in this slice

- `convert-agreement-to-format` tool (still calls REST via `makeApiRequest` until slice 2b wraps the real template engine as a service)
- `trigger-agreement` tool (same reason)
- REST `/agreements` route rewire (slice 3, once #211 lands and the `crud.ts` collision surface is clear)

## Validation

- `npm test` in `server/`: 10/10 suites pass, 125 tests total (4 new)
- `npx tsc --noEmit` clean
- Convert + trigger paths untouched, so the `TemplateArchiveProcessor` integration in `server/handlers/agreements.ts` is not affected

## Author Checklist

- [x] DCO sign-off provided
- [x] Tests added and passing
- [x] Typecheck clean
- [x] No transport-specific imports in `services/`
- [x] `AgreementNotFoundError` used for typed errors (no bare `throw new Error(...)` from the service)

### DIFF
--- a/server/db/client.ts
+++ b/server/db/client.ts
@@ -1,0 +1,8 @@
+import type { PostgresJsDatabase } from 'drizzle-orm/postgres-js';
+import * as schema from './schema';
+
+// The RI wires the shared Drizzle handle onto `res.locals.db` at request time
+// (see server/index.ts). Services take this typed handle as their first arg so
+// they stay transport-agnostic — the MCP handler and REST routes can both call
+// the same service function without going through an internal HTTP loop.
+export type Database = PostgresJsDatabase<typeof schema>;
--- a/server/handlers/mcp.ts
+++ b/server/handlers/mcp.ts
@@ -17,6 +17,7 @@ import {
     UpstreamApiError,
 } from '../services/errors';
 import { listTemplates, getTemplateById } from '../services/templateService';
+import { listAgreements, getAgreementById } from '../services/agreementService';
 import type { Database } from '../db/client';
 
 const HOST = process.env.HOST || 'localhost';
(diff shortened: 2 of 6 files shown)


niallroche (changes requested):
Same pagination note as #211: listAgreements(db) is an unbounded db.select().from(Agreement). The apap://agreements resource was implicitly capped at ≤100 via parseQueryParams; it now returns the full table. Suggest the same { limit, offset } bounds on the primitive so slice 3's REST list route inherits paging rather than slicing in memory. The resolveAgreementRuntime helper and the templateHash / resource: URI handling look faithful to the old inline logic.

PR #214, opened 2026-07-18
feat(services): port convertAgreement to agreementService, rewire REST + MCP (slice 2b)

JayDS22:
## Summary

Slice 2b of the shared-service-layer port from [apap-mcp-poc](https://github.com/JayDS22/apap-mcp-poc). Adds `convertAgreement` to `agreementService.ts` so both the REST `/agreements/:id/convert/:format` route and the MCP `convert-agreement-to-format` tool call the same function. Eliminates one more internal `makeApiRequest` HTTP loop from `handlers/mcp.ts`.

## Stack

Stacked on top of #213 (slice 2), which is stacked on #211 (slice 1). Base is `main`; the diff will collapse to the three files below once #211 and #213 land.

## Scope reduction from the originally planned slice 2b

`triggerAgreement` extraction is deferred to **slice 2c**. When the trigger logic runs behind a service-layer import chain, jest cannot propagate `--experimental-vm-modules` through the dynamic imports `@accordproject/template-engine` performs, and the existing trigger tests fail with `ERR_VM_DYNAMIC_IMPORT_CALLBACK_MISSING_FLAG`. The old inline trigger path (unchanged in this slice) works because template-engine loads from the same module context as the test file. Solving that needs its own investigation and does not belong in this slice.

Convert does not exhibit the same behavior — its code path does not hit the dynamic import chain in the same way, and all existing tests pass with it moved behind the service.

### `server/services/agreementService.ts`
- Adds `resolveAgreementRuntime` (private helper): loads agreement, resolves associated template by hash or URI, reconstructs the Cicero template archive. Mirrors the inline `resolveAgreement` in `handlers/agreements.ts`.
- Adds `convertAgreement(db, agreementId, format)`: uses `resolveAgreementRuntime` + `TemplateArchiveProcessor.draft`. Throws `AgreementNotFoundError` on missing agreement, `AgreementConversionError` on template resolution or render failure.
- `templatebuilder` is imported from `../handlers/templatebuilder`. It is technically under `handlers/` in the current tree but is a transport-agnostic utility (no Express or MCP SDK imports). Moving it under a proper utility directory is a follow-up refactor.

### `server/handlers/agreements.ts`
- REST `/:id/convert/:format` now calls `convertAgreement` service. Adds the same strict-numeric `:id` guard as #208.
- Trigger route unchanged (slice 2c territory).
- Inline `resolveAgreement` helper kept because trigger still uses it. Duplication with `resolveAgreementRuntime` is intentional for slice 2b and gets removed in slice 2c.

### `server/handlers/mcp.ts`
- MCP `convert-agreement-to-format` tool now calls `convertAgreement` service directly. Adds the same strict-numeric guard.
- Module fn `draftAgreement` deleted (was the HTTP-loop wrapper).
- Trigger tool and its module fn `triggerAgreement` unchanged.

## Not in this slice

- `triggerAgreement` extraction (slice 2c, once the jest+VM issue is understood)
- REST route rewire for the list + CRUD endpoints (slice 3, once #211/#213 land)

## Validation

- `npm test` in `server/`: 10/10 suites pass, 125 tests total
- `npx tsc --noEmit` clean
- Trigger path completely untouched; existing trigger tests (including the ones that exercise template-engine dynamic imports) all still pass

## Author Checklist

- [x] DCO sign-off provided
- [x] Tests passing (no new tests: `convertAgreement` is exercised end-to-end by the existing REST route, and MCP tool coverage is a slice-3 topic)
- [x] Typecheck clean
- [x] Transport-agnostic service (no Express or MCP SDK imports in `services/`)

### DIFF
--- a/server/db/client.ts
+++ b/server/db/client.ts
@@ -1,0 +1,8 @@
+import type { PostgresJsDatabase } from 'drizzle-orm/postgres-js';
+import * as schema from './schema';
+
+// The RI wires the shared Drizzle handle onto `res.locals.db` at request time
+// (see server/index.ts). Services take this typed handle as their first arg so
+// they stay transport-agnostic — the MCP handler and REST routes can both call
+// the same service function without going through an internal HTTP loop.
+export type Database = PostgresJsDatabase<typeof schema>;
--- a/server/handlers/agreements.ts
+++ b/server/handlers/agreements.ts
@@ -8,6 +8,7 @@ import { TemplateArchiveProcessor } from '@accordproject/template-engine';
 import { HttpTemplateRetriever } from './retrievers/HttpTemplateRetriever';
 import { Template as CiceroTemplate } from '@accordproject/cicero-core';
 import { AgreementNotFoundError } from '../services/errors';
+import { convertAgreement } from '../services/agreementService';
 import { asyncHandler } from '../middleware/errorHandler';
 
 /**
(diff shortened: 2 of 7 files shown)


niallroche (commented):
This convert extraction looks faithful to the old REST path, and keeping trigger out of this slice was the right scope choice.

One small thing I’d like to see, either here or in the next service-layer cleanup: direct service-level coverage for `convertAgreement`. At the moment the new service boundary is covered mostly indirectly through the existing REST route tests. Since `resolveAgreementRuntime` and `convertAgreement` will be shared by both REST and MCP, a focused test for success plus one template-resolution or draft-failure path would make the new boundary easier to trust.

Not a blocker from me, especially with #211/#213 carrying the pagination fixes, but worth pinning while this code is being moved into the shared service layer.

niallroche (approved):

PR #216, opened 2026-07-21
feat(services): port triggerAgreement, unify REST + MCP trigger path (slice 2c)

JayDS22:
## Summary

Slice 2c of the shared-service-layer port. Closes the runtime-half extraction that slice 2b started with convert. `triggerAgreement` now lives in `agreementService.ts` alongside `convertAgreement`, and both the REST `/agreements/:id/trigger` route and the MCP `trigger-agreement` tool call the same function. Two more internal `makeApiRequest` HTTP loops are gone from `handlers/mcp.ts`, and the inline `resolveAgreement` helper in `handlers/agreements.ts` is deleted (both REST routes now use the service's private `resolveAgreementRuntime`).

## Stack

Stacked on top of #214 (slice 2b), which is stacked on #213 (slice 2), which is stacked on #211 (slice 1). Base is `main`; the diff will collapse to the three files below once the earlier slices land.

## About the "slice 2c blocker" from 2026-07-18

I closed slice 2b with a note that trigger extraction was blocked by a jest+VM dynamic-import issue and needed its own investigation. **That turned out to be a false alarm.** In yesterday's debug loop I was running `npx jest handlers/agreements.test.ts` on the failing tests, which strips the `--experimental-vm-modules` flag that the `test` script sets via `node --experimental-vm-modules node_modules/.bin/jest`. The moment I switched back to `npm test`, all 125 tests passed with trigger fully extracted. No code-side workaround needed.

### `server/services/agreementService.ts`
- Adds `triggerAgreement(db, agreementId, requestBody)`. Uses `resolveAgreementRuntime` + `TemplateArchiveProcessor`. Throws `InvalidPayloadError` on `$class` mismatch, `ValidationError` on Concerto validation failure, `AgreementTriggerError` on execution failure. Wraps init + trigger in a single try so runtime failures surface uniformly.
- `resolveAgreementRuntime` template-not-found path now throws a plain `Error` so `globalErrorHandler` renders `{ error: message }` at 500 — this preserves the exact wire shape the REST test suite already asserts. Typed error usage would have changed the response body.

### `server/handlers/agreements.ts`
- REST `/:id/trigger` calls the service. Catches the three "recoverable" typed error families and maps them to the legacy `{ isError: true, errorMessage, errorDetails }` shape at HTTP 200 for backward compatibility:
  - `InvalidPayloadError` -> full message on both fields
  - `ValidationError` -> hardcoded top-line + first concerto error message (via `err.details.errors[0].message`)
  - `AgreementTriggerError` -> `err.upstreamMessage` unwrapped (strips the typed `"Failed to trigger agreement N: "` prefix so existing clients see the raw business error unchanged)
- Deletes the inline `resolveAgreement` helper (unused now).
- Drops `templateFromDatabase`, `TemplateArchiveProcessor`, and the duplicate error-import block.

### `server/handlers/mcp.ts`
- MCP `trigger-agreement` tool calls the service directly. Parses the JSON string payload that the MCP tool contract expects. Adds the same strict-numeric guard as slice 1/2/2b.
- Deletes the module-level `triggerAgreement` HTTP-loop wrapper.
- Imports the service function as `triggerAgreementService` to avoid a name collision with the MCP tool name.

## Validation

- `npm test` in `server/`: 10/10 suites pass, 125 tests total
- `npx tsc --noEmit` clean
- Wire shape for REST trigger errors preserved: existing `{ isError, errorMessage, errorDetails }` clients see exactly the same body strings

## Author Checklist

- [x] DCO sign-off provided
- [x] Tests passing
- [x] Typecheck clean
- [x] Backward-compatible REST wire shape preserved
- [x] Transport-agnostic service (no Express or MCP SDK imports in `services/`)

### DIFF
--- a/server/db/client.ts
+++ b/server/db/client.ts
@@ -1,0 +1,8 @@
+import type { PostgresJsDatabase } from 'drizzle-orm/postgres-js';
+import * as schema from './schema';
+
+// The RI wires the shared Drizzle handle onto `res.locals.db` at request time
+// (see server/index.ts). Services take this typed handle as their first arg so
+// they stay transport-agnostic — the MCP handler and REST routes can both call
+// the same service function without going through an internal HTTP loop.
+export type Database = PostgresJsDatabase<typeof schema>;
--- a/server/handlers/agreements.ts
+++ b/server/handlers/agreements.ts
@@ -2,59 +2,18 @@ import express from 'express';
 import { Agreement, AgreementInsertSchema, Template as DbTemplate } from '../db/schema';
 import { buildCrudRouter } from './crud';
 import { concertoValidation } from './concertovalidation';
-import { templateFromDatabase, extractTemplateForDatabase } from './templatebuilder';
+import { extractTemplateForDatabase } from './templatebuilder';
 import { eq } from 'drizzle-orm';
-import { TemplateArchiveProcessor } from '@accordproject/template-engine';
 import { HttpTemplateRetriever } from './retrievers/HttpTemplateRetriever';
 import { Template as CiceroTemplate } from '@accordproject/cicero-core';
-import { AgreementNotFoundError } from '../services/errors';
-import { convertAgreement } from '../services/agreementService';
+import {
+    AgreementNotFoundError,
+    AgreementTriggerError,
+    InvalidPayloadError,
+    ValidationError,
+} from '../services/errors';
+import { convertAgreement, triggerAgreement } from '../services/agreementService';
 import { asyncHandler } from '../middleware/errorHandler';
-
-/**
- * @param db The database handle stored on `res.locals.db`.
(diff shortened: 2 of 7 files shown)


niallroche (commented):
The REST + MCP unification is clean and preserving the legacy { isError: true } shape at HTTP 200 is the right backward-compat call. One thing worth tightening:

In triggerAgreement, the persist step spreads the whole row:

await db.update(Agreement).set({ ...agreement, state: triggerResult.state })...

That writes every column back — including the primary key — on what is really a single-field state transition. It's unnecessary write surface and opens a concurrent-update clobber (any column that changed between the read and this write gets overwritten with the stale value). A targeted set({ state: triggerResult.state }) is safer and says exactly what it does. Minor — not a blocker for the slice.

niallroche (approved):

Issue #221, opened 2026-07-27
MCP SDK upgrade


niallroche:
Tracks the MCP SDK upgrade and the dependency bumps deferred from dependabot #219. The security-only subset of #219 landed via #220; this issue captures the rest, which are held pending code changes and the spec cutover.

## Deferred bumps (from #219)

| Package | From → proposed | Disposition |
|---|---|---|
| `@modelcontextprotocol/sdk` | 1.15.1 → 1.29.0 | **Don't take 1.29** — migrate to the 2.0 split packages instead (below) |
| `@accordproject/template-engine` | 2.7.1 → 2.9.0 | Fold into the SDK migration pass |
| `nodemon` | 1.18.10 → 3.1.14 | Dev dep, major ×2 — batch with the migration |
| `wait-on` | 3.2.0 → 9.1.0 | Dev dep, major ×6 — batch with the migration |

## Why not just take SDK 1.29

1. **It breaks the build.** 1.15 → 1.29 tightened the `.tool()` generics; the `z.enum(['html','markdown'])` schema on the convert tool in `mcp.ts` tips past the TS depth limit and fails `tsc` with `TS2589` (the three `z.string()` tools stay under). So 1.29 needs an `mcp.ts` schema refactor just to land.
2. **It doesn't reach the RC.** 1.29.x is the *old* release line. Support for the 2026-07-28 RC ships in a new split package family, published Jul 21: `@modelcontextprotocol/{core,server,node,express,hono,fastify,codemod}@2.0.0-beta.5` (`@modelcontextprotocol/sdk@2.0.0-beta.5` → 404).

Doing the `mcp.ts` refactor for 1.29 now and again for 2.0 in ~1–2 weeks is duplicated work.

## Plan

- **Now:** stay on SDK 1.15; security patches covered by #220.
- **After Jul 28 finalises:** read the final spec, then migrate `sdk` → the `@2.0.0-beta` split packages directly (`core` + `server` + the express binding), doing the `mcp.ts` schema refactor **once** against the final target. Fold `template-engine` 2.9 and the dev-dep majors into the same pass.
- APAP is insulated on the wire (production uses `McpServer`). One POC test assertion needs updating for the post-RC diff: spec PR modelcontextprotocol/modelcontextprotocol#3002 moves `serverInfo` into `result._meta['io.modelcontextprotocol/serverInfo']` and demotes envelope `clientInfo` to SHOULD — only matters for hand-rolled init responses, which we don't have.

## References

[...]

PR #223, opened 2026-07-29
feat(errors): standardize MCP JSON-RPC codes into -32020..-32099 range

JayDS22:
## Summary

The MCP spec reserves `-32020..-32099` for implementation-defined error codes. Previously the `ServiceError -> McpError` mapping in `serviceErrorToResourceError` was a coarse 2-way split (404 -> `InvalidParams`, else -> `InternalError`), losing per-subclass identity.

This threads a `jsonRpcCode: number` through the `ServiceError` hierarchy so each subclass carries its own MCP code:

| Subclass | Code |
|---|---|
| `TemplateNotFoundError` | `-32020` |
| `AgreementNotFoundError` | `-32021` |
| `TemplateDuplicateError` | `-32022` |
| `AgreementConversionError` | `-32023` |
| `AgreementTriggerError` | `-32024` |
| `InvalidPayloadError` | `-32025` |
| `ValidationError` | `-32026` |
| `UpstreamApiError` | `-32027` |
| `ServiceError` (bare default) | `-32028` |

MCP clients can now branch on the integer code without parsing message strings; the string `code` field (e.g. `TEMPLATE_NOT_FOUND`) stays as the machine-readable form for REST + MCP consumers.

## What this PR does

- Adds `MCP_ERROR_CODES` central constant table in `services/errors.ts`, with tests asserting all values sit inside the reserved range and no two subclasses collide on the same code.
- Threads `jsonRpcCode` through every `ServiceError` subclass constructor.
- Updates `serviceErrorToResourceError` and `serviceErrorToCallToolResult` in `handlers/mcp.ts` to emit the typed code instead of the coarse 2-way split.
- Extends `services/errors.test.ts` with per-subclass code assertions + range-invariant tests.

## Stack

Base of a three-PR chain that closes Proposal Core and aligns MCP with the 2026-07-28 spec finalisation:

1. **This PR** - JSON-RPC error range standardisation
2. Subscriptions/listen SEP-2575 upstream port (adds `SubscriptionInvalidUriError -32029` in this range)
3. Slice 3 REST list unification (closes Proposal Core service-layer port)

Merge earliest first; later branches will rebase forward.

## Validation

- `npm ci` in `server/`: clean
- `npm run build`: clean (`tsc -p .`)
- `npm test`: 10 suites / 153 tests, all pass (was 141 pre-branch, +12 from `errors.test.ts`)

## Related

Sets the JSON-RPC code convention that Issue **#221** ("MCP SDK upgrade") builds on for the 2.0-beta.5 migration.

## Author Checklist

- [x] Branch rebased on latest `main`

### DIFF
--- a/server/handlers/mcp.test.ts
+++ b/server/handlers/mcp.test.ts
@@ -30,12 +30,13 @@ import mcpRouter, {
     SERVER_INSTRUCTIONS,
     CACHE_HINTS,
 } from './mcp';
-import { McpError, ErrorCode } from '@modelcontextprotocol/sdk/types.js';
+import { McpError } from '@modelcontextprotocol/sdk/types.js';
 import {
     AgreementConversionError,
     AgreementNotFoundError,
     ServiceError,
     TemplateNotFoundError,
+    UpstreamApiError,
 } from '../services/errors';
 import { InMemoryTransport } from '@modelcontextprotocol/sdk/inMemory.js';
 import { Client } from '@modelcontextprotocol/sdk/client/index.js';
@@ -224,12 +225,14 @@ describe('mcp typed error helpers', () => {
     });
 
     describe('serviceErrorToResourceError', () => {
-        it('returns an McpError with InvalidParams code for 404-style ServiceErrors', () => {
+        it('returns an McpError whose code is the ServiceError subclass jsonRpcCode', () => {
             const err = new TemplateNotFoundError('tmpl-1');
             const wrapped = serviceErrorToResourceError(err);
 
             expect(wrapped).toBeInstanceOf(McpError);
-            expect(wrapped.code).toBe(ErrorCode.InvalidParams);
+            // TemplateNotFoundError sits at MCP_ERROR_CODES.TEMPLATE_NOT_FOUND (-32020),
+            // in the -32020..-32099 range reserved for implementation-defined MCP errors.
+            expect(wrapped.code).toBe(-32020);
             expect(wrapped.message).toContain('tmpl-1');
 
             const data = wrapped.data as { error: { code: string; message: string; details?: unknown } };
@@ -237,12 +240,13 @@ describe('mcp typed error helpers', () => {
             expect(data.error.message).toContain('tmpl-1');
         });
 
-        it('uses InternalError code for non-404 ServiceErrors', () => {
-            const err = new ServiceError('CUSTOM', 500, 'kaboom', { reason: 'overflow' });
+        it('propagates whatever jsonRpcCode a bare ServiceError carries', () => {
+            // Base-class throw with an explicit jsonRpcCode in the -32020..-32099 range.
+            const err = new ServiceError('CUSTOM', 500, -32050, 'kaboom', { reason: 'overflow' });
             const wrapped = serviceErrorToResourceError(err);
 
             expect(wrapped).toBeInstanceOf(McpError);
-            expect(wrapped.code).toBe(ErrorCode.InternalError);
+            expect(wrapped.code).toBe(-32050);
             // McpError prepends "MCP error <code>:" to the constructor message; check substring.
             expect(wrapped.message).toContain('kaboom');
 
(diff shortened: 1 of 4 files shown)


niallroche:
This is a clear improvement over the coarse 404→InvalidParams / else→InternalError mapping — typed, specific codes are the right direction, and `-32020..-32099` sits correctly inside JSON-RPC's `-32000..-32099` "reserved for implementation-defined server errors" band, so the range is spec-legal. The collision/range tests in `errors.test.ts` are a nice touch for keeping it that way.

Three things worth a look before merge:

**1. This is a wire-visible change — call it out.** Any MCP client currently branching on the numeric `code` will see different values: a bare `ServiceError` that used to surface as `-32603` (InternalError) now comes back as `-32028`. That's fine for a pre-1.0 MCP surface, but it deserves a line in the PR body / changelog so downstream client authors aren't surprised.

**2. Confirm the fallback for *non-*`ServiceError` throws.** I can see `MCP_ERROR_CODES.SERVICE_ERROR = -32028` as the fallback for a bare `ServiceError`, but I couldn't see `serviceErrorToResourceError` / `serviceErrorToCallToolResult` in `errors.ts` — they're in `mcp.ts`. Can you confirm that something thrown that *isn't* a `ServiceError` at all (a raw `Error`, a Concerto exception) still maps to `-32603` InternalError, and doesn't fall through to `-32028`? `-32028` should mean "a ServiceError we didn't specialize," not "anything unexpected," otherwise you lose the "this was genuinely internal" signal.

**3. Standard codes where they fit?** `INVALID_PAYLOAD` (-32025) and `VALIDATION_ERROR` (-32026) map conceptually onto JSON-RPC's predefined `-32602` Invalid params. Using custom codes uniformly is a defensible choice (one namespace, more specific), but a generic JSON-RPC client won't recognize `-32025` as "bad params" the way it would `-32602`. Worth a sentence on the rationale — I'd lean toward keeping the custom codes for specificity, just documenting that they carry the same `data.code` string so clients have a stable non-numeric key.

Note: for `UpstreamApiError` (-32027), consider carrying the upstream HTTP status in `error.data` — a `502` from an upstream is diagnostically different from a `404`, and the numeric JSON-RPC code alone flattens that.

JayDS22:
Amended `990704f` addressing the three review notes:

**1. Wire-visibility change flagged.** Confirmed: any MCP client that was branching on the numeric `code` from a bare `ServiceError` will see `-32028` instead of the old `-32603`. Subclassed errors (which is everything the RI actually throws today) already carried typed strings via `error.code`, so the `code` string field stays a stable non-numeric key across the change. Called this out here so downstream client authors have a heads-up.

**2. Non-`ServiceError` fallback confirmed.** The `-32028` code applies ONLY to callers that explicitly construct a bare `ServiceError`. Raw `Error`, Concerto exceptions, and every other throw path still bubble unwrapped through resource handlers (`getAgreement`, `getTemplates`, etc. all guard with `if (err instanceof ServiceError)` before wrapping), so the SDK default handler maps them to `-32603 InternalError` as before. Made this contract explicit in the `serviceErrorToResourceError` JSDoc (new "Invariant" paragraph in `mcp.ts`) so a future editor can't accidentally widen the wrapper.

**3. `UpstreamApiError` carries HTTP status in `error.data`.** Already threaded through: `UpstreamApiError` constructor passes `{ upstreamUrl, httpStatus, upstreamBody }` into `details`, `ServiceError.toJSON()` spreads `details` into the serialized payload, and `serviceErrorToResourceError` writes `error.toJSON()` into the `McpError.data` field. Added a targeted test in `mcp.test.ts` (`'UpstreamApiError surfaces httpStatus in error.data.details for diagnostic parity with the REST 502'`) so this contract is pinned against regression.

On the `INVALID_PAYLOAD` / `VALIDATION_ERROR` vs `-32602` observation: keeping the custom `-32025` / `-32026` codes for specificity per your lean. The `data.error.code` string (`INVALID_PAYLOAD` / `VALIDATION_ERROR`) is the stable non-numeric key clients can branch on if they want the generic "bad params" bucket.

`npm test`: 10 suites / 154 tests, all pass.

JayDS22:
Closing this as superseded. **#227** (MCP SDK 2.0 migration) landed the `ProtocolError` hierarchy with the coarse 404 -> `INVALID_PARAMS`, else -> `INTERNAL_ERROR` split that this PR was going to replace with per-subclass typed `jsonRpcCode` dispatch.

The typed-code refinement (each `ServiceError` subclass carrying its own code in the reserved `-32020..-32099` range) is still a valid improvement, but not load-bearing: clients that want to branch on error type already have `error.data.error.code` (string like `TEMPLATE_NOT_FOUND`) as a stable non-numeric key, and the coarse numeric split satisfies the JSON-RPC layer's needs.

Leaving this closed rather than reopening as a follow-up; can revisit if a specific client actually needs per-subclass numeric dispatch. The dependency PRs **#224** (subscriptions/listen) and **#225** (slice 3 REST list unification) will be rewritten against post-#227 main and reopened / amended as needed.

PR #224, opened 2026-07-29
feat(mcp): port subscriptions/listen SEP-2575 preview upstream from POC

JayDS22:
## Summary

Ports the SEP-2575 subscribe-to-resource-updates preview from `apap-mcp-poc` into the upstream RI. Wire-compatible with the 2026-07-28 RC's `subscriptions/listen` shape (single POST, `uris[]` param, returns `{ subscriptionId }`, notifications on the same session's transport as `notifications/resources/updated` tagged with `io.modelcontextprotocol/subscriptionId`).

## What this PR does

- **`server/services/subscriptionRegistry.ts`**: new. Session -> URI -> Subscriber index with `O(subscribers)` `notify()` and `clearSession()`. Fire-and-forget: broken send callbacks are swallowed so `notify()` cannot cascade back to a DB write path.
- **`server/services/errors.ts`**: adds `SubscriptionInvalidUriError` (`-32029`, in the reserved MCP range from the base PR).
- **`server/handlers/mcp.ts`**: `getServer()` accepts an optional `getSessionId` callback. When provided, registers the `subscriptions/listen` handler on the underlying `Server` and advertises `{ resources: { subscribe: true } }` in server capabilities. Both transport close hooks (Streamable HTTP `transport.onclose` and SSE `res.on("close")`) call `subscriptionRegistry.clearSession(sid)` to prevent per-session leaks.

## Stack

Middle of a three-PR chain. **Depends on #223** (JSON-RPC error range) — the new `SubscriptionInvalidUriError -32029` sits in that reserved range.

Sequence:

1. **#223** - JSON-RPC error range standardisation (merge first)
2. **This PR** - subscriptions/listen SEP-2575 upstream port
3. Slice 3 REST list unification (opens after this merges or against this branch)

## Not in this slice (follow-ups)

- `notify()` call sites in `templateService` / `agreementService` writes. Subscribers can register today but nothing fires yet; that plumbing lands per-service so each write path can be reviewed on its own.
- Migration to the SDK's native subscription primitive when the v2 beta line (`@modelcontextprotocol/{core,server,...}@2.0.0-beta.x`) exposes it. The registry contract does not change. Tracked under Issue **#221**.

## Validation

- `npm run build`: clean
- `npm test`: 11 suites / 172 tests (+19 registry tests: happy path, `notify()` fan-out, session isolation, `clearSession()` semantics, error-swallow guarantee), all pass

## Related

Refs Issue **#221** (MCP SDK upgrade — subscriptions primitive migration path in 2.0-beta.5).

## Author Checklist

- [x] Rebased on top of #223

### DIFF
--- a/server/handlers/mcp.ts
+++ b/server/handlers/mcp.ts
@@ -21,7 +21,13 @@ import {
     AgreementConversionError,
     AgreementTriggerError,
     UpstreamApiError,
+    SubscriptionInvalidUriError,
 } from '../services/errors';
+import {
+    subscriptionRegistry,
+    isValidResourceUri,
+    NotificationPayload,
+} from '../services/subscriptionRegistry';
 import { listTemplates, getTemplateById } from '../services/templateService';
 import {
     listAgreements,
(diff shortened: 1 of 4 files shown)


niallroche:
The registry design is clean — dual `bySession` / `byUri` index for O(subscribers) fan-out, idempotent `subscribe()`, fire-and-forget `notify()` that swallows send-callback errors so a slow subscriber can't cascade back into a write path, and `clearSession()` wired to both `transport.onclose` and the SSE `res.on("close")`. Lifecycle hygiene is genuinely good and the "nothing fires yet" framing is honest.

That said, because this is the moment the design gets set, two gaps are much cheaper to close now than after `notify()` is wired:

**1. No authorisation on `subscribe()` — this is the one to resolve before notifications are wired.** Right now any session can subscribe to any `apap://templates/{id}` or `apap://agreements/{id}`, and the only gate is the URI-prefix scheme check. That's latent while nothing fires, but the instant `notify()` is connected to the write paths it becomes a cross-tenant information leak: a session receives update notifications — and, depending on payload, content — for resources it could never `read`. This RI has real multi-tenant auth (org scoping / RLS on the enterprise side), so `subscribe()` needs to enforce the *same* authorisation as a resource read, at registration time. Even if you merge this as scaffolding, I'd bake in the check (or at minimum a `// SECURITY: auth required before notify() is wired` guard that fails closed) so it can't be forgotten when the fun part lands.

**2. Relatedly, decide the notification payload shape now.** If `notify()` carries resource *data*, it compounds (1) and duplicates the auth surface. If it carries only "`uri` changed, re-fetch," the client's normal `resources/read` re-read re-enforces auth for free. I'd strongly prefer the latter — thin notifications, authoritative re-read — and it's worth locking in before consumers build against a data-carrying shape.

**3. No bound on subscriptions.** There's no per-session or global cap on distinct URIs, so `bySession`/`byUri` grow unbounded with subscribe traffic — a memory-exhaustion vector, especially if the subscription path ends up reachable without auth. A per-session cap (and ideally a global one) with a `SubscriptionLimitError` would close it.

Minor: confirm the session id feeding `clearSession(sid)` / `subscribe` is unguessable and bound to the authenticated session — the whole isolation model rests on `sid` not being forgeable or reusable across clients.

Happy to see this land as a preview once (1) has a decision — scaffolding is fine, but I'd treat "wire `notify()` without the auth check" as the actual vulnerability, so better to encode the requirement in code now.

JayDS22:
Amended `04678bd` addressing all three review notes:

**1. Fail-closed auth guard baked in.** Added a prominent `SECURITY` comment block at the top of the `subscriptions/listen` handler spelling out the invariant: the handler MUST NOT be wired to `notify()` until it enforces the same authorisation as `resources/read` for the target URI. Contract for the follow-up PR that wires notify() is spelled out in-code (auth check before `subscribe()`, thin payload only, transport-bound unguessable `sessionId`). The whole point is to make the requirement unmissable when the fun part lands.

**2. Notification payload is thin, locking in.** Explicitly documented that `NotificationPayload` carries `uri` + `subscriptionId` only, never resource data. Client receives "uri changed, re-fetch" and re-reads via `resources/read`, which re-enforces auth. The `notify()` doc block now says do not widen the payload shape.

**3. Per-session cap + `SubscriptionLimitError` added.**
- New `MAX_SUBSCRIPTIONS_PER_SESSION = 100` constant exported from `subscriptionRegistry.ts`. Enforced in `subscribe()` BEFORE the unsubscribe-and-replace step, so re-subscribing an existing URI is a no-op for the cap (only new URIs count).
- New `SubscriptionLimitError` (code `-32030`, HTTP 429, in the `-32020..-32099` range from **#223**) carrying `{ sessionId, currentCount, limit }` so clients get the boundary in one round-trip.
- Registry's obsolete "avoids importing errors.ts" comment removed: errors.ts has no imports of its own, so no cycle exists.
- Handler-side: `subscribe()` throws are now caught and wrapped via `serviceErrorToResourceError` so both `SubscriptionInvalidUriError -32029` and `SubscriptionLimitError -32030` surface as proper `McpError`s carrying their assigned codes. Non-`ServiceError` still bubbles for the SDK to map to `-32603`.

On the `sessionId` unguessability point: `getSessionId()` returns the transport's own session identifier (Streamable HTTP: SDK-generated per `initialize`; SSE: the request-scoped id set up in `createServer`). Client cannot forge or reuse across sessions.

Five new tests in `subscriptionRegistry.test.ts` covering: cap enforced on new URIs, cap not tripped by re-subscribes, caps are per-session (session A unaffected by session B hitting cap), `SubscriptionLimitError` payload shape.

`npm test`: 11 suites / 178 tests, all pass.

dselman:
Resolve conflicts please

JayDS22:
Force-pushed `e2ae328` (was `04678bd`). Rewritten against post-**#227** main:

- Reset to `origin/main`; the old branch was built on SDK 1.x + assumed **#223** as base, both no longer accurate.
- Dropped the **#223**-adjacent `MCP_ERROR_CODES` table (**#223** closed as superseded, per its close comment).
- Uses SDK 2.0 `setRequestHandler(method: string, schema, handler)` shape (was `(schema, handler)` in SDK 1.x).
- SSE path removed alongside the transport it depended on; only the Streamable HTTP `onclose` calls `subscriptionRegistry.clearSession(sid)`.
- `ServiceError` subclasses wrap via main's existing `serviceErrorToResourceError` -> `ProtocolError`; no new error-hierarchy work smuggled into this PR.

**Validation:** `npm run build` clean, `npm test` 11 suites / 181 tests all pass (+24 from main baseline for registry + cap tests).

Ready for review again.

JayDS22:
Closing this without landing. A pre-review pass caught two blocking wire-conformance issues that make the current shape wrong to merge:

**1. `SubscribeParamsSchema` is fabricated.** SDK 2.0 already exports `SubscriptionsListenRequestParamsSchema` from `@modelcontextprotocol/core` with:

```ts
{
  _meta: { progressToken?, "io.modelcontextprotocol/related-task"? },
  notifications: {
    toolsListChanged?: boolean,
    promptsListChanged?: boolean,
    resourcesListChanged?: boolean,
    resourceSubscriptions?: string[]
  }
}
```

This PR uses `{ uris: string[] }` at params root, which is non-conformant. Spec-compliant clients would fail params validation.

**2. The custom `setRequestHandler + subscriptionRegistry` duplicates a native SDK 2.0 primitive.** `@modelcontextprotocol/server` exports `McpHttpHandler` with `notify: ServerNotifier`, plus `SUBSCRIPTION_ID_META_KEY`, `SubscriptionsListenResult`, `SubscriptionFilter`, `InMemoryServerEventBus`. The subscription registry + custom handler here reinvents ~450 lines of what the SDK provides. Also the notification payload puts `io_modelcontextprotocol_subscriptionId` at `params` root, but the wire spec puts `"io.modelcontextprotocol/subscriptionId"` (dotted, with slash) inside `_meta`.

**Approach going forward:** rather than amend this PR to fix the schema shape only (throwaway work once the SDK-primitive migration lands), closing and opening a fresh PR that uses `McpHttpHandler.notify.resourceUpdated` directly. Follow-up issue filed to track the correct-shape work.

The design work isn't wasted:
- The `subscriptionRegistry` transport-agnostic surface, per-session cap, and `clearSession` wiring pattern all inform the follow-up.
- The SECURITY fail-closed guard doc (auth check contract before `notify()` is wired) transfers directly.
- The 24 tests around cap boundaries and cross-session isolation give the follow-up a template.

Closing rather than deferring to keep the review queue clean.

JayDS22:
Follow-up issue filed: **#232** ([feat(mcp): implement subscriptions/listen using SDK 2.0 native McpHttpHandler](https://github.com/accordproject/apap/issues/232)). Scope, target APIs, and what transfers from this PR are documented there.

Issue #217, opened 2026-07-21
[follow-up] Expose paged reads via MCP resource URIs for apap://templates and apap://agreements


JayDS22:
The service primitives `listTemplates(db, { limit, offset })` and `listAgreements(db, { limit, offset })` (added in #211 and #213 respectively) support paged reads with the same ≤100 cap the REST layer already enforces via `parseQueryParams`. The MCP resource callbacks currently call them with no opts (defaults 100/0), so MCP clients can only ever see the first 100 rows on `apap://templates` and `apap://agreements`.

## Motivation

- The `ttlMs` / `cacheScope` hints from #201 currently annotate a bounded but non-paged payload; a client that needs the 101st row cannot get it via the resource path.
- REST callers coming through slice 3's unification will get proper paging; MCP callers should have parity.
- Not urgent: the ≤100 default matches the current REST behaviour, so this isn't a regression — it's a missing capability.

## Options to consider

1. **Query-string form on the resource URI**: `apap://templates?limit=50&offset=100`. Check whether the MCP SDK's `ResourceTemplate` can surface query-string vars alongside path vars. Cleanest for existing clients that already handle URI variables.
2. **Separate paged tool**: e.g. `listTemplatesPaged(cursor?)` returning `{ items, nextCursor }`. More explicit but forks the resource surface.
3. **Client-driven cursor via ETag/If-None-Match**: unusual for MCP, probably out of scope.

## References

- Service primitives with clamped bounds: #211 (templates), #213 (agreements)
- MCP resource callbacks call the primitives with no opts on the head branches of #211 / #213 (`handlers/mcp.ts`, `getTemplates` / `getAgreements` module fns and their `apap://templates/{templateId}` / `apap://agreements/{agreementId}` list callbacks)
- Cache hint annotations that motivated the primitive-side clamp: #201 (`CACHE_HINTS` in `handlers/mcp.ts`)

Track under slice 3 (REST unification) or as a separate slice 4 depending on whether the design converges on option 1 or option 2.

JayDS22:
Pinning the paging shape here so it does not drift while the service-layer PRs land.

[...]

PR #225, opened 2026-07-29
feat(crud): slice 3 REST list delegates to service via listService

JayDS22:
## Summary

Extends the `buildCrudRouter` helper with an optional `listService` param. When provided, the `GET /` handler delegates the DB read + count to that service function instead of running its own inline query. The router still owns query-string parsing (`parseQueryParams`), `whereClause` construction (`defaultWhereClause` with `SAFE_IDENTIFIER_RX` / operator whitelist), and `orderClause` construction; the service owns the DB select + count.

This gives REST and MCP a single code path for reads, matching the pattern already in place for the single-resource operations (`getTemplateById`, `getAgreementById`). **Closes the Proposal Core service-layer port on the code side.**

## What this PR does

- **`server/services/templateService.ts`**: new `listTemplatesPaged(db, opts)` returning `{ items, total }`. Accepts `whereClause` + `orderClause` from the caller so REST filter/sort semantics stay uniform. `limit` clamped 1..100, `offset` clamped >= 0.
- **`server/services/agreementService.ts`**: new `listAgreementsPaged` with the same shape.
- **`server/handlers/crud.ts`**: adds `ListService<TRow>` type + `listService?` option on `CrudRouterOptions`. `GET /` branches on whether `listService` is provided; the pre-slice-3 inline path is preserved as the fallback so resources without a service layer (`sharedmodels`) keep working unchanged.
- **`server/handlers/templates.ts`, `agreements.ts`**: pass `listService` so their REST list routes now call the service functions.

## Stack

Top of a three-PR chain. **Depends on #224** (subscriptions/listen) which depends on **#223** (JSON-RPC error range).

Sequence:

1. **#223** - JSON-RPC error range (merge first)
2. **#224** - subscriptions/listen SEP-2575 preview
3. **This PR** - slice 3 REST list unification (merge last, closes Proposal Core)

## Unlocks

- MCP resource paths and REST `GET /` share one implementation for list. Behaviour changes to one propagate to both.
- The `{ items, total }` return shape is the piece Issue **#217** needs for paged `apap://templates` and `apap://agreements` MCP resource URIs.

## Not in this slice (follow-ups)

- `sharedmodels` service layer + `listSharedModelsPaged` (its service file does not exist yet; `sharedmodels` stays on the default DB path).
- MCP resource handlers switching from `listTemplates` (array) to `listTemplatesPaged` (`{items,total}`) to expose `total` in `ReadResourceResult`. Tracked under **#217**.

## Validation

- `npm run build`: clean
- `npm test`: 11 suites / 172 tests, all pass. `templates.test.ts` and `agreements.test.ts` continue to exercise the list route through the new delegation path.

## Related

- Refs Issue **#217** (MCP resource-URI paging) - this PR delivers the `{items,total}` primitive; the resource-handler switch is a small follow-up.
- Closes the service-layer port workstream (last inline DB read in `buildCrudRouter` now delegates).

## Author Checklist

- [x] Rebased on top of #224 (which is on top of #223)

### DIFF
--- a/server/handlers/agreements.ts
+++ b/server/handlers/agreements.ts
@@ -12,7 +12,7 @@ import {
     InvalidPayloadError,
     ValidationError,
 } from '../services/errors';
-import { convertAgreement, triggerAgreement } from '../services/agreementService';
+import { convertAgreement, triggerAgreement, listAgreementsPaged } from '../services/agreementService';
 import { asyncHandler } from '../middleware/errorHandler';
 
 const router = express.Router();
@@ -88,7 +88,8 @@ const crudRouter = buildCrudRouter({
     typeName: 'Agreement',
     // ponytail: cast schema to any due to zod v3 -> v4 upgrade depth-instantiation
     // issue with drizzle-zod. Runtime unaffected.
-    validateBody: { schema: AgreementInsertSchema as any, custom: (body) => concertoValidation('Agreement', body) }
+    validateBody: { schema: AgreementInsertSchema as any, custom: (body) => concertoValidation('Agreement', body) },
+    listService: (db, opts) => listAgreementsPaged(db, opts),
 });
 
 /**
--- a/server/handlers/crud.ts
+++ b/server/handlers/crud.ts
@@ -186,6 +186,21 @@ export type InsertValidator = {
     custom?: (body:any) => Promise<ValidationResult>,
 }
 
+/**
+ * Signature for a service-layer list function that the router can delegate
+ * to. Returns `{ items, total }`; the router wraps that into the full
+ * PaginatedResponse envelope with page / limit / totalPages metadata.
+ */
+export type ListService<TRow> = (
+    db: any,
+    opts: {
+        whereClause?: SQL;
+        orderClause?: SQLWrapper | null;
+        limit: number;
+        offset: number;
+    },
+) => Promise<{ items: TRow[]; total: number }>;
+
 interface CrudRouterOptions<T extends PgTable<any> & TableWithId> {
     table: T;
     typeName: string;
@@ -193,6 +208,15 @@ interface CrudRouterOptions<T extends PgTable<any> & TableWithId> {
     validateBody?: InsertValidator;
     transformResponse?: (item: any) => any;
     transformRequest?: (req: Request) => any;
+    /**
+     * Optional service-layer list function. When provided, the GET / handler
+     * delegates DB read + count to it instead of the inline path. Router
+     * still owns query-string parsing (`parseQueryParams`), `whereClause`
+     * construction (`defaultWhereClause` with `SAFE_IDENTIFIER_RX` /
+     * operator whitelist), and `orderClause` construction; the service owns
+     * the DB select + count. Keeps REST and MCP on one code path for reads.
+     */
+    listService?: ListService<any>;
 }
 
 function getSingleQueryParam(value: unknown): string | undefined {
@@ -292,7 +316,8 @@ export function buildCrudRouter<T extends PgTable<any> & TableWithId>({
     buildWhereClause,
     validateBody,
     transformResponse,
-    transformRequest
+    transformRequest,
+    listService,
 }: CrudRouterOptions<T>): Router {
     const router = Router();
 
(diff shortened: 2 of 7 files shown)


niallroche (commented):
The split is right: router keeps query parsing, `whereClause` (with `SAFE_IDENTIFIER_RX` + operator whitelist), and `orderClause`; the service owns select + count and clamps `limit` to `1..100` and `offset` to `>= 0`. Clamping in the service is the correct home even if the router also does it — defense in depth against a future caller. Nice.

A few things:

**1. Pagination needs a deterministic total order — confirm `orderClause` is never effectively null.** In `listTemplatesPaged` the ordering is applied conditionally (`opts.orderClause ? baseQuery.orderBy(...) : baseQuery`). If a request can reach the service with no `orderClause`, the DB returns rows in unspecified order, and `limit`/`offset` paging over an unordered set can repeat or skip rows across pages. A list endpoint with offset pagination should always fall back to a stable key (primary key or `createdAt`). Can you confirm the router always supplies a default order, and if not, default it inside the service so paging is stable by construction?

**2. `count` and `select` are separate queries — note the read skew.** `total` and `items` are fetched in two round-trips with no shared snapshot, so under concurrent writes a client can see `total: 11` with 10 items (or vice versa). Usually acceptable for list endpoints; just worth a comment, and if you want them consistent, a `repeatable read` transaction around the pair does it.

**3. The REST/MCP divergence cuts against the PR's stated goal.** REST now returns `{ items, total }` while the MCP resource path still returns a bare array via `listTemplates`. Deferring the MCP switch to #217 is reasonable, but since the headline is "unify REST + MCP," this actually *widens* the gap until the follow-up lands. Please make sure #217 is tracked and the temporary shape difference is documented so a client dev doesn't assume `total` exists on both.

**4. Test the boundaries.** I don't see coverage for the clamping/edge cases the service now owns: `limit=0` and `limit=1000` (→ 100), negative `offset`, and `offset` past `total` (expect empty `items`, correct `total`). These are exactly the paths that regress silently.

Nit: the retained inline fallback for `sharedmodels` means that resource keeps the old path *without* the service's clamping/consistency. Fine as transitional, but worth a `// TODO(#217?)` so it doesn't quietly become permanent.

JayDS22:
Amended `ed91aad` addressing all four review notes + the nit:

**1. Default `orderClause` fallback for pagination determinism.** Both `listTemplatesPaged` and `listAgreementsPaged` now default to `asc(Template.id)` / `asc(Agreement.id)` when no `orderClause` is provided, so `limit`/`offset` paging is stable by construction. Caller-supplied `orderClause` is honoured as before. Documented in the service function bodies (line-comment right above the fallback).

**2. Read-skew note added in comment.** Called out in both services that `count` and the row-fetch are separate queries without a shared snapshot, so `total` and `items.length` can disagree by a row or two under concurrent writes. Kept the current two-query pattern (acceptable for list endpoints), with the `repeatable read` transaction escape hatch documented for anyone who ever needs strict consistency.

**3. REST/MCP shape divergence documented.** The `apap://templates` / `apap://agreements` MCP resource handlers still call the pre-slice-3 `listTemplates` / `listAgreements` (bare array). Slice 3 delivers the `{ items, total }` primitive that **#217** needs; the resource-handler switch to `listTemplatesPaged` / `listAgreementsPaged` is a small follow-up on that issue. Called out explicitly in the amendment so a client dev seeing the mixed shape has the reason.

**4. Boundary tests added.** Seven new tests in `templateService.test.ts` covering `limit=0` (clamped to 1), `limit=1000` (clamped to 100), negative offset (clamped to 0), in-range pass-through, default orderClause is applied (pagination determinism), caller-provided orderClause is honoured, and `total` surfaces from the count query. Same clamp logic in `listAgreementsPaged` is symmetric; happy to add mirror tests there if you'd prefer both sides pinned.

**Nit: `sharedmodels` inline fallback.** Added a `TODO(#217)` block-comment above the fallback branch in `crud.ts` noting that the fallback path does NOT get the clamping / default-order / read-skew safeguards, and that once a `sharedModelService.ts` + `listSharedModelsPaged` land, the fallback branch can be deleted and `listService` made required on `CrudRouterOptions`.

`npm test`: 11 suites / 185 tests, all pass.

dselman:
Resolve conflicts please

JayDS22:
Force-pushed `d08c47e` (was `ed91aad`). Rewritten against post-**#227** main:

- Reset to `origin/main`; the old branch stacked on **#223** + **#224** work that is either now closed or handled separately (**#224** rewritten in parallel).
- Only the slice-3 delta: `listTemplatesPaged` / `listAgreementsPaged` in the service files, `ListService` type + `listService?` option in `buildCrudRouter`, wire-through in templates.ts / agreements.ts.
- Kept main's `as any` casts on the drizzle-zod schemas (post-#227 pattern) — no scope creep into that.
- `sharedmodels` inline fallback preserved with a `TODO(#217)` marker so it can be removed once `sharedModelService.ts` lands.
- Confirmed envelope contract: service returns `{ items, total }`; router wraps into main's existing `{ items, total, page, limit, totalPages }`. No wire-observable shape change.

**Validation:** `npm run build` clean, `npm test` 10 suites / 159 tests all pass (unchanged from main baseline).

**Not in this PR** (deliberate, follow-up):
- Your **#226** test-coverage gaps (limit-clamp end-to-end, page=0/-1, 23505 -> 400, POST body assertion) — folding into a small test-only PR after this merges to keep the diff focused.

Ready for review.

JayDS22:
Amendment `bc7c608` pushed — adds the 12 tests for the safeguards the paged variants introduce (6 for `listTemplatesPaged`, 6 symmetric for `listAgreementsPaged`).

Coverage added:
- `clamps limit to 100` when caller requests more, and `>= 1` when they request less
- `clamps offset >= 0` when negative
- `applies a default orderClause when caller passes null` (pagination determinism gate)
- `honours a caller-provided orderClause` (verifies the fallback does not overwrite)
- `surfaces total from the count-query destructure`

Extended the shared `createMockDb` in both service test files with `orderBy: jest.fn().mockReturnThis()` so the fluent chain does not throw when the paged variant appends `.orderBy(...)` to the row query.

Deliberate scope note: skipped a router-integration test asserting `templatesRouter` forwards clauses to `listTemplatesPaged` verbatim. The wiring is a two-line closure and a mock-based test would exercise jest's mock system rather than delegation semantics. Real coverage of that regression needs a live-Postgres integration test which is a separate PR shape (would want it running under the postgres-18-smoke job from **#231**).

Validation: `npm test` 10 suites / 171 tests, all pass.

niallroche:
still conflicts present

JayDS22:
Rebased on `a74867e` (post #237 + #242). Conflicts were import-only in `handlers/templates.ts` and `services/templateService.ts` (unioned the imports; slice-3 additions preserved). 107/107 tests pass locally across the affected suites (`templateService`, `agreementService`, `crud`, `templates`, `agreements`). Typecheck clean. Ready for re-check.

niallroche (approved):

PR #165, opened 2026-04-14
test: add unit tests for ```CRUD``` router

Thomas-Sedhom:
### Replaces #133 (closed by stale bot)

Added a full unit test suite for the buildCrudRouter utility in crud.ts. The tests cover all five HTTP methods the router exposes and use a mock DB layer so there's no real database dependency.
What's covered

GET /

- Returns paginated results and correctly caps limit at 100 even when a higher value is passed
- Enriches user items with an empty roles array when typeName is users

POST /
- Runs schema validation, custom validation, transformRequest, and transformResponse in the right order and with the right data
- Returns 400 with field-level details when schema validation fails
- Returns 400 with custom error messages when custom validation fails
- Returns 409 on duplicate key constraint violations (pg error code 23505)

GET /:id
- Returns a single item passed through transformResponse
- Returns 404 when the record doesn't exist

PUT /:id
- Updates the record and returns it transformed
- Returns 409 on unique constraint violation

DELETE /:id
- Deletes the record and returns { status: 'deleted' }

Screenshot 2026-03-19 205113
Screenshot 2026-03-19 205102

PR #226, opened 2026-07-31
test(crud): add GET list envelope + GET :id + POST happy-path coverage (takeover of #165)

JayDS22:
## Summary

Adds unit test coverage for `server/handlers/crud.ts` surface not exercised by the existing 26 tests in `server/handlers/crud.test.ts`. The router's list envelope shape (`items` + `total` + `page` + `limit` + `totalPages`), single-fetch happy + 404, and POST insert happy path were all untested.

## Attribution

Original framing and coverage-gap identification: **[#165](https://github.com/accordproject/apap/pull/165)** by @Thomas-Sedhom (opened 2026-04-14). That PR sat without author activity for ~4 months while `server/handlers/crud.ts` was hardened with:

- SQL-injection guards (`bdc536e`)
- Strict numeric-ID parsing (**#208**, `eff6134`)
- 404-on-missing DELETE (**#180**)
- Pagination guards (**#177**)
- Pagination-query parsing hardening (`5e6782d`)

A straight rebase of **#165** would test a router that no longer exists; replacing rather than rebasing per the maintainer stale-PR takeover policy. Direction and problem statement are @Thomas-Sedhom's; the diff here targets the current router shape.

## What this PR does

7 new tests across 3 describe blocks:

**`GET / list route (envelope + pagination shape)`**
- returns PaginatedResponse envelope with `items` + `total` + `page` + `limit` + `totalPages`
- returns an empty items array + `total=0` when the table has no rows
- paginates past total: page beyond last returns empty items but preserves `total`
- computes `totalPages` as `ceil(total / limit)`

**`GET /:id single-fetch route`**
- returns 200 with the row when the id exists
- returns 404 when the id does not exist

**`POST / happy path`**
- inserts a valid body and returns 200 with the created row

No changes to `crud.ts` itself; test-only.

## Implementation notes

- Mock helper for the list route uses a queued `then` interceptor on the db mock so the two sequential awaits (count query, items query) can each return their own value without brittle `mockResolvedValueOnce` chaining on individual chain methods.
- `globalErrorHandler` is now wired into the test apps for the new describe blocks so thrown errors surface as JSON responses that assertions can inspect, matching the runtime shape at `server/index.ts:77`.

## Validation

- `npm ci` in `server/`: clean
- `npm run build`: clean
- `npm test`: 10 suites / 158 tests, all pass (was 141 pre-branch, +17 from this PR + prior main advances)

## Related

- Refs **#165** (original coverage-gap framing, being superseded by this PR; will close **#165** once this lands)
- Related hardening already on `main`: **#208**, **#180**, **#177**

## Author Checklist

### DIFF
--- a/server/handlers/crud.test.ts
+++ b/server/handlers/crud.test.ts
@@ -7,6 +7,7 @@ import * as db from './templatebuilder';
 import * as validationModule from './concertovalidation';
 import { parseQueryParams } from './crud';
 import { TemplateInsertSchema } from '../db/schema';
+import { globalErrorHandler } from '../middleware/errorHandler';
 
 jest.mock('../db/schema');
 jest.mock('./templatebuilder');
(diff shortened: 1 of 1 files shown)


dselman (approved):

niallroche:
Good addition of the  CRUD tests (takeover of #165)** — *clean, credits the original author, but misses the exact #225 boundaries it exists to lock down.* Envelope shape is pinned well and it covers empty-table / paginate-past-total. 
Gaps: 
1. **no end-to-end limit-clamp assertion** — `?limit=1000→100` is only tested at the `parseQueryParams` unit level, and the lower clamp `?limit=0→1` isn't tested anywhere; 
2. no `?page=0/-1` clamp test; (3) the **23505 duplicate→400** branch is untested; (4) POST happy-path asserts the mock was called but never asserts the returned body. 

Also note its envelope is `{items,total,page,limit,totalPages}` — richer than the `{items,total}` I'd described for #225, so confirm the final contract with yourself on #225.

PR #228, opened 2026-07-31
fix(crud): add `~` prefix for case-insensitive filter values (closes #125)

### LINKED ISSUE #125: get template along with  author parameter,case sensitive issue
Scenarion :- if a person has given his name as "ROB" now when he later like viewing his template if he types "rob"
or uses like "Rob" then the template won't be shown so 
made the author name parameter case insensitive
solved here :-
https://github.com/accordproject/apap/pull/124

ex through images :- 
before :- 

Image

after :- 

Image
worked both for " Purushotham "and "purushotham "

Thank you 🦖

JayDS22:
Closes **#125**.

## Summary

Adds a case-insensitive filter operator to the CRUD router's `defaultWhereClause` via a `~` prefix on the query-string value. Example:

```
GET /templates?author=~Rob
```

matches templates authored by `"rob"`, `"ROB"`, `"Rob"`, etc. by mapping to Postgres `ILIKE`.

**The bug (#125):** a template created with author `"Purushotham"` was not findable via `?author=purushotham` because the router built a case-sensitive equality clause.

## Design choice (why this shape, not #124)

Adds the case-insensitive branch to the existing per-value operator scheme (`>=`, `<=`, `!=`, `>`, `<`) rather than special-casing the `author` field. This keeps the router generic, opt-in per query, and discoverable alongside the other operators.

**#124** (the prior fix attempt by @FRIERIN) was closed unmerged because it hard-coded `author` handling into the generic router, which does not scale to other text fields with the same problem (`displayName`, `description`, `uri`, etc.). This PR fixes the same intent generically.

Existing case-sensitive equality queries continue to work unchanged (no wire-observable regression).

## Security

- Column identifier safety is still enforced by `SAFE_IDENTIFIER_RX` before any `sql.raw()` interpolation (no change to that guard).
- The `ILIKE` operand is passed as a **bound parameter**, not interpolated, so `%` and `_` wildcards from client input are handled safely by Drizzle's parameter binding — no SQL injection risk introduced.

## What this PR does

- `server/handlers/crud.ts`: adds a case-insensitive branch in `defaultWhereClause` that matches `value` starting with `~`, slices off the prefix, and emits `ILIKE ${operand}` as a parameterised SQL fragment. Placed **before** the standard operator match so `~` never collides with the numeric comparison operators.

## Tests

3 new tests in the SQL-injection-prevention describe block:

- `?author=~Rob` matches without crashing (happy path)
- `?author=~` with empty operand does not crash (edge case)
- `?author=~%25admin%25` (ILIKE wildcard metacharacters) is safely passed through parameter binding

## Validation

- `npm run build`: clean
- `npm test`: **10 suites / 154 tests**, all pass (was 151 pre-branch, +3 from this PR)

## Attribution

Original bug reporter: **@FRIERIN** in **#125** (with the \"Purushotham\" example screenshots). The prior fix attempt **#124** was author-specific; this generalises the same intent across all text columns.

## Related

- Closes **#125**
- Supersedes **#124** (closed unmerged); credit to @FRIERIN for identifying the bug

## Author Checklist

### DIFF
--- a/server/handlers/crud.test.ts
+++ b/server/handlers/crud.test.ts
@@ -437,6 +437,35 @@ describe('SQL injection prevention in defaultWhereClause', () => {
             .get('/templates?author=null');
         expect(res.status).not.toBe(500);
     });
+
+    // Case-insensitive filter via `~` prefix (closes #125). Original bug: a
+    // template created by "Purushotham" was not findable by `?author=purushotham`
+    // or `?author=Purushotham` because the router built a case-sensitive
+    // equality clause. The `~` prefix opts the value into an ILIKE comparison,
+    // which Postgres treats as case-insensitive. Backwards compatible: existing
+    // queries without the prefix keep the case-sensitive equality behaviour.
+    it('does not crash when ~ prefix requests a case-insensitive match', async () => {
+        const res = await request(app)
+            .get('/templates?author=~Rob');
+        expect(res.status).not.toBe(500);
+    });
+
+    it('does not crash when ~ prefix is used with an empty operand', async () => {
+        // `?author=~` after the tilde is an empty value; the router should
+        // still handle it (Postgres ILIKE against an empty string is a valid
+        // query, matches empty string columns only). The point is no crash.
+        const res = await request(app)
+            .get('/templates?author=~');
+        expect(res.status).not.toBe(500);
+    });
+
+    it('does not crash when ~ prefix contains SQL-like metacharacters', async () => {
+        // ILIKE treats % and _ as wildcards. Passing them as bound parameters
+        // is safe (no SQL injection), but the router should still not crash.
+        const res = await request(app)
+            .get('/templates?author=~%25admin%25');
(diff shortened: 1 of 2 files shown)


dselman (approved):

niallroche:
1. `~` case-insensitive filter** — *injection fundamentals sound, but three should-fixes.* Value is a bound parameter (no SQLi) and the `~` branch can't smuggle operators/identifiers. 

2. **`%` and `_` aren't escaped** — `?author=~%` matches everything; it's advertised as equality but silently becomes pattern-match. Fix: escape metacharacters + `ESCAPE '\'`, or use `lower(col)=lower(val)`; 

3. the **test named "SQL-like metacharacters" actually certifies the bug** — it only asserts `not 500`, so `~%admin%` "passes" *because* wildcards are live; (3) **`~` on a non-text column 500s** (`?id=~5` → Postgres type error), and the `hasOwnProperty` guard is skipped when `table` is absent.

PR #229, opened 2026-07-31
feat: add support for Postgres 18 (closes #118)

JayDS22:
Closes **#118**.

Bumps `server/compose.yaml` from `postgres:17-alpine` to `postgres:18-alpine` and updates the volume mount to match Postgres 18's new PGDATA layout.

## What changed

- **`server/compose.yaml`**: image `postgres:17-alpine` → `postgres:18-alpine`, volume `db-data:/var/lib/postgresql/data` → `db-data:/var/lib/postgresql` (Postgres 18 relocated PGDATA to `/var/lib/postgresql/<MAJOR>/docker` and now declares the `VOLUME` one level up). Inline comment documents the incompatibility with existing PG 17 volumes.
- **`server/upgrade-postgres-17-to-18.sh`** (new, `chmod +x`): scripted `pg_dumpall` + volume recreate + restore path for developers with an existing `server_db-data` volume they want to preserve.
- **`server/README.md`**: new "Upgrading from Postgres 17" section covering both the throw-away-the-dev-volume path and the scripted dump/restore path.

## Attribution

Rebases **@Vaggelis-Arg**'s original **#119** (commit `f67f81e`), which was auto-closed by the stale bot before it could land despite @dselman's *"This looks great"* review. Per stale-PR takeover policy, opening as a replacement with co-authorship credit preserved via `Co-authored-by:` trailer on the commit.

Script content is byte-identical to Vaggelis's original. Added the README section and a compose.yaml pointer to the script that **#119** did not include.

## Migration impact

Existing dev environments with a `server_db-data` volume created under Postgres 17 will NOT auto-migrate: the mount path change means PG 18 sees an empty volume and initialises a fresh cluster inside a subdirectory the old data is not in. README documents both remediation paths.

Fresh installs (no existing volume) just work: `docker compose up` on a new machine boots PG 18 cleanly, and `npx drizzle-kit push` bootstraps the schema as before.

## Validation

- `docker compose -f server/compose.yaml config` parses clean
- `docker compose -f server/compose.yaml build` succeeds against current `main` (verified locally 2026-07-31)
- `postgres-js@3.4.5` and `drizzle-orm@0.45.2` both work with PG 18 per the manual test transcript in **#119**

## Author Checklist

- [x] DCO sign-off on every commit
- [x] Co-authorship credit for @Vaggelis-Arg via commit trailer
- [x] Rebased on latest `main`
- [x] Docker build validated locally
- [x] No em-dashes in commit message or PR body

PR #227, opened 2026-07-31
feat(mcp): migrate to @modelcontextprotocol/{core,server,express,node,client}@2.0.0 (closes #221)

JayDS22:
Closes **#221**.

## Summary

Migrates the RI off `@modelcontextprotocol/sdk@1.15.1` (now at `1.30.0` latest, both on the deprecated 1.x line) to the split-package family at **`2.0.0` final**. **`2.0.0` was published as stable while #221 was open**, so this skips both the 1.29 hop and the 2.0-beta hop that #221 originally planned around and goes straight to the stable line.

## Load-bearing outcome

**`TS2589` count = 0.** The type-instantiation-depth error on the `convert-agreement-to-format` tool schema (`z.enum(['html','markdown'])`) that blocks any SDK 1.x ≥ 1.29 is **gone** under 2.0.0. Local build + test suite both clean.

## Package changes (`server/package.json`)

- **Removed:** `@modelcontextprotocol/sdk@1.15.1`
- **Added:** `@modelcontextprotocol/core@2.0.0` (types, schemas, protocol constants)
- **Added:** `@modelcontextprotocol/server@2.0.0` (`McpServer`, `ResourceTemplate`, `InMemoryTransport`, `EventStore`, error hierarchy)
- **Added:** `@modelcontextprotocol/express@2.0.0`
- **Added:** `@modelcontextprotocol/node@2.0.0` (`NodeStreamableHTTPServerTransport`)
- **Added:** `@modelcontextprotocol/client@2.0.0` (test-side `Client`, was a subpath of the old SDK)
- **Bumped:** `@accordproject/template-engine 2.7.1 → 2.9.0`, `nodemon 1.18 → 3.1` (dev), `wait-on 3.2 → 9.1` (dev)
- **Bumped:** `zod 3.24.3 → 4.4.3` (required by SDK 2.0 at runtime: *\"Raw-shape inputSchema fields must be Zod v4 schemas\"*)
- **Bumped:** `drizzle-zod 0.7.1 → 0.8.3` (zod v4 compat)

### `server/handlers/inmemoryeventstore.ts`
Off private `@modelcontextprotocol/sdk/dist/esm/*` import paths (a pre-existing bomb flagged in the #221 issue body) onto public `@modelcontextprotocol/server`. `EventStore` + `JSONRPCMessage` are now public exports; no runtime logic changed.

### `server/handlers/mcp.ts`
[...]

PR #231, opened 2026-08-06
chore(cleanup): follow-ups to #227 / #228 / #229 review comments

JayDS22:
Addresses @niallroche's post-merge follow-ups on the Aug 5 batch. Grouped into one PR because each item is small and unrelated to the others.

## #227 cleanup (`server/package.json`)

- Remove `"undefined": "^0.1.0"` from dependencies. Junk package accidentally installed during the SDK migration spike; not referenced anywhere in code.
- Pin `@accordproject/template-engine` to `4.0.0` exact. Was caret-ranged (`^4.0.0`), which silently allows any 4.x minor bump and did not match the PR-body description that mentioned 2.9.
- Pin `@modelcontextprotocol/{core,server,express,node,client}` to `2.0.0` exact (were caret-ranged). SDK 2.x line is new enough that a minor bump could break; explicit pin matches house convention.
- Pin `zod` to `4.4.3` exact (was caret-ranged).
- **Not fixed here:** the zod v4.4.3 / drizzle-zod 0.7.1 peer relationship is loose because drizzle-zod's peerDep advertises zod `^3`. Runtime works via the depth-workaround already documented at buildCrudRouter call sites; a drizzle-zod bump to a zod-v4-aware release is a separate PR when upstream cuts one.

## #228 cleanup (`server/handlers/crud.ts` + `crud.test.ts`)

- Escape ILIKE metacharacters (`%` `_` `\`) in the `~` prefix operand and add `ESCAPE '\\'`. **Before:** `?author=~%admin%` silently became a wildcard contains-match despite the operator advertising case-insensitive EQUALITY. **After:** literal `"%admin%"` is matched as-is; wildcard behaviour would need a separate operator.
- Rewrite the "SQL-like metacharacters" test that previously certified the bug (only asserted not-500, so `~%admin%` "passed" because wildcards were live). The renamed test documents the escape contract explicitly.
- **Fail-closed hasOwnProperty guard:** switched `if (table && !hasOwnProperty(table, key))` to `if (!table || !hasOwnProperty(table, key))` so callers that omit the table argument now filter every key out, rather than silently interpolating arbitrary user-supplied keys.
- **Not fixed here** (TODO in code): `~` on a non-text column (`?id=~5`) still surfaces as a Postgres type error. Needs a shared column-metadata accessor on the router; scope creep for this cleanup.

## #229 cleanup (`.github/workflows/build.yml` + `server/upgrade-postgres-17-to-18.sh`)

- **New `postgres-18-smoke` job in build.yml:** standalone Linux-runner job with a `postgres:18-alpine` service, applies the app schema via `npx drizzle-kit push`. Cannot attach services to the matrix build (they only work on Linux runners); this standalone job is the minimal way to make "tested against PG18" true in CI, not just in local docker compose.
- Add `[ -s "$DUMP" ]` guard after `pg_dumpall` in the upgrade script. `pg_dumpall` can exit 0 even when the dump is empty or truncated; the guard refuses to proceed to volume-delete unless the dump has content, closing the silent-data-loss window.
- Un-hardcode `VOLUME_NAME`: accept as first positional arg or `VOLUME_NAME` env var, default remains `server_db-data` for backward compatibility. Fixes the break under custom docker-compose project names.

## Validation

- `npm run build`: clean
- `npm test`: 10 suites / 159 tests, all pass

## Related

- Follow-ups to @niallroche's Aug 5 review comments on **#227** / **#228** / **#229**
- **#226** test-coverage gaps (limit-clamp E2E, page=0/-1, 23505 → 400, POST body assertion) are deliberately NOT included here; folding into a test-only PR that lands alongside or after **#225** to keep this diff focused

## Author Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -56,6 +56,42 @@ jobs:
           name: npm-logs
           path: C:\npm\cache\_logs\
 
+  # Postgres 18 smoke check. Standalone job (services: only run on Linux
+  # runners, cannot be attached to the matrix build above). Proves the pinned
+  # postgres:18-alpine image stands up cleanly and accepts the app's schema
+  # via drizzle-kit push. Follow-up to #229 review: "tested against PG18"
+  # is now true in CI, not just in local docker compose.
+  postgres-18-smoke:
+    name: Postgres 18 schema smoke
+    runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgres:18-alpine
+        env:
+          POSTGRES_PASSWORD: 1baddeed
+          POSTGRES_DB: postgres
+        options: >-
+          --health-cmd "pg_isready -U postgres"
+          --health-interval 5s
+          --health-timeout 5s
+          --health-retries 10
+        ports:
+          - 5432:5432
+    steps:
+      - uses: actions/checkout@v4
+      - name: Use Node.js 22.x
+        uses: actions/setup-node@v4
+        with:
+          node-version: 22.x
+      - name: Install server deps
+        working-directory: server
+        run: npm ci
+      - name: Apply schema to Postgres 18 via drizzle-kit push
+        working-directory: server
+        env:
+          POSTGRES_URL: postgres://postgres:1baddeed@localhost:5432/postgres
+        run: npx drizzle-kit push --config drizzle.config.ts
+
   publish:
     needs:
       - build
--- a/server/handlers/crud.test.ts
+++ b/server/handlers/crud.test.ts
@@ -363,19 +363,28 @@ describe('POST without validateBody', () => {
 
 describe('SQL injection prevention in defaultWhereClause', () => {
     let app: express.Application;
+    // Captures each SQL clause passed to db.where(...). The first call is
+    // the count query, the second the row query; both receive the same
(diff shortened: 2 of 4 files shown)


niallroche:
package issue to be resolved before merge

niallroche (approved):
Confirmed the two that mattered: `"undefined"` is gone from `server/package.json` and `@accordproject/template-engine` is pinned exact `4.0.0`. The zod-v4 / drizzle-zod deferral is the right call — don't force it ahead of an upstream drizzle-zod release. Ok to merge.

Two things worth folding in while you're here (neither blocking):

1. **`postgres-18-smoke` only applies the schema.** It stands up `postgres:18-alpine` and runs `drizzle-kit push`, which proves PG18 accepts the schema — but not that RLS still isolates rows on 18, which was the actual risk. One authenticated query through `set_config('app.user_id', …)` / `app.org_id` would turn this from a migration check into a real isolation check.
2. **The MCP SDK packages are still `^2.0.0` (caret).** Since this PR is the pin-exact cleanup, pin `@modelcontextprotocol/*` to an exact 2.x that includes the UriTemplate ReDoS fixes (#965, #1334) — #243's resource read path calls `match()` on client-controlled URIs, so the floor matters.

PR #236, opened 2026-08-07
chore(): Upgrade cicero-core to 2.x, concerto-core/util to 4.x

mttrbrts:
## Summary

- Bump `@accordproject/cicero-core` in `server/` from `0.25.2` to the latest 2.x release, `2.1.1`.
- Bump `@accordproject/concerto-core` and `@accordproject/concerto-util` from `3.21.0`/`3.20.4` to the latest `4.1.5`, matching what `@accordproject/template-engine@4.0.0` already nests as a transitive dependency (previously the RI carried two separate major versions of concerto-core side by side).
- `@accordproject/template-engine` is left at `^4.0.0`, which already resolves to the latest `4.0.0`.

### Fallout fixed

- `concertovalidation.ts` and `scripts/drizzle-gen.ts` both constructed `ModelManager` with a `strict: true` option; that option was removed in concerto-core 4.x (versioned namespaces are already mandatory), so the object literal no longer type-checks. Dropped it.
- `scripts/drizzle-gen.ts`'s visitor callbacks (`getAllDeclarations`, `getProperties`, `getOwnProperties`) relied on concerto-core's introspect API returning concretely-typed arrays for parameter inference. In concerto-core 4.x these getters are typed `any`, so the callback parameters needed explicit `any` annotations to satisfy `noImplicitAny`.
- `Template#toArchive` in cicero-core 2.x makes the `language` parameter required (still handled as falsy at runtime) — updated the one test call site.
- `Template.fromArchive` now enforces the `accordproject.cicero` semver range declared in a template's `package.json` against the running cicero-core version. Bumped the test fixture's declared range (`server/test/archives/latedeliveryandpenalty-typescript/package.json` and the paired `apap_request.json`) from `^0.25.0` to `^2.0.0`.
- cicero-core 2.x requires Node `>=22`. Bumped `server/package.json` `engines.node` and the `Dockerfile`'s `ARG NODE_VERSION` from `20.18.1` to `22.13.1` — CI (`.github/workflows/build.yml`) already runs the matrix on Node 22.x/24.x, so this aligns the local dev/Docker story with CI rather than changing it.
- Updated `server/README.md`'s example request bodies from `"cicero": "0.25.x"` to `"cicero": "2.x"`.

## Test plan

- [x] `npm run build` (`tsc -p .`) passes cleanly in `server/`.
- [x] `npm test` in `server/`: 152/159 passing. The remaining 7 failures are all in `handlers/agreements.test.ts` and share one root cause — a test fixture template imports Concerto models over HTTPS from `models.accordproject.org`, which this sandbox's network policy blocks (`403` at the proxy). Verified these same 7 tests fail identically against the pre-upgrade code in this same sandbox, so this is a pre-existing sandbox network limitation, not a regression — these should pass in the project's actual CI, which has normal internet access.
- [x] Confirmed `db/schema.ts` (generated from `model/protocol.cto` via `scripts/drizzle-gen.ts`) is unaffected — the model itself didn't change, and a diff against a pre-upgrade copy is identical.

---
_Generated by [Claude Code](https://claude.ai/code/session_0126thR7JHJE8j7ogxv4cGLj)_

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -6,7 +6,7 @@
 
 # Want to help us make this template better? Share your feedback here: https://forms.gle/ybq9Krt8jtBL3iCk7
 
-ARG NODE_VERSION=20.18.1
+ARG NODE_VERSION=22.13.1
 
 ################################################################################
 # Use node image for base image for all stages.
--- a/server/README.md
+++ b/server/README.md
@@ -162,7 +162,7 @@ curl --request POST \
 		"$class": "org.accordproject.protocol@1.0.0.TemplateMetadata",
 		"runtime": "typescript",
 		"template": "clause",
-		"cicero": "0.25.x"
+		"cicero": "2.x"
 	},
 	"logo": null,
 	"templateModel": {
@@ -209,7 +209,7 @@ Response:
 		"$class": "org.accordproject.protocol@1.0.0.TemplateMetadata",
 		"runtime": "typescript",
 		"template": "clause",
-		"cicero": "0.25.x"
+		"cicero": "2.x"
 	},
 	"logo": null,
 	"templateModel": {
@@ -258,7 +258,7 @@ Response:
 		"$class": "org.accordproject.protocol@1.0.0.TemplateMetadata",
 		"runtime": "typescript",
 		"template": "clause",
-		"cicero": "0.25.x"
+		"cicero": "2.x"
 	},
 	"logo": null,
 	"templateModel": {
@@ -311,7 +311,7 @@ Response:
 				"$class": "org.accordproject.protocol@1.0.0.TemplateMetadata",
 				"runtime": "typescript",
 				"template": "clause",
-				"cicero": "0.25.x"
+				"cicero": "2.x"
 			},
 			"logo": null,
 			"templateModel": {
@@ -385,7 +385,7 @@ Response:
 		"$class": "org.accordproject.protocol@1.0.0.TemplateMetadata",
 		"runtime": "typescript",
 		"template": "clause",
-		"cicero": "0.25.x"
+		"cicero": "2.x"
 	},
 	"logo": null,
 	"templateModel": {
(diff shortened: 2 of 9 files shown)


JayDS22 (approved):
Overall this looks good and the PR body is thorough. CI is clean across all platforms. Two minor things:

**`toArchive(undefined)` in the test (`agreements.test.ts:82`)**

The comment says "save the archive in the template's own declared runtime (typescript)" but then passes `undefined` to satisfy the required parameter. If cicero-core 2.x added a required `language` param, passing `undefined` is just deferring the type complaint to runtime. The test fixture already declares `runtime: typescript` in its `package.json`, so why not pass `'typescript'` explicitly? That matches the comment's stated intent and is more obviously correct.

**`strict: true` removal (`concertovalidation.ts:95`, `drizzle-gen.ts:1581`)**

Both sites drop the option with no explanation at the call site. Someone hitting this in six months won't know if it was deliberate or accidental. A short comment noting the option was removed in concerto-core 4.x would save that future investigation.

Nothing blocking. Happy to approve once the `toArchive` call is either clarified or switched to the explicit `'typescript'` value.

JayDS22:
Following up: noticed \`3fc868a\` landed the explicit \`toArchive('typescript')\` change two minutes after I posted the approve, so my review referenced the un-fixed state. Fix is in, thanks.

The \`strict: true\` removal comment note is still worth adding when convenient (concertovalidation.ts + drizzle-gen.ts), but not blocking.

mttrbrts:
The `strict: true` removal comment is already in `3fc868a` — same commit as the `toArchive('typescript')` fix, at `concertovalidation.ts:22` and `drizzle-gen.ts:258`. Nothing further to do here.

---
_Generated by [Claude Code](https://claude.ai/code)_

mttrbrts:
Thanks for the review @JayDS22 !

PR #240, opened 2026-08-09
feat(server): freeze signed/completed records; fix SIGNNG typo

mttrbrts:
## Summary

- Fixes the `SIGNNG` → `SIGNING` typo in `AgreementStatusType` at the source (`model/protocol.cto`) and propagates it through every generated artifact: `openapi.json`, `client/typescript/apap.ts`, `index.md`/`_site`, and the Drizzle schema/DB enum (`server/db/schema.ts`), plus a migration (`server/drizzle/0002_fix_signing_enum_typo.sql`) that uses `ALTER TYPE ... RENAME VALUE` so existing rows with `SIGNNG` keep their data instead of failing a cast.
- Freezes an agreement's **deal terms (`data`) as soon as signing starts**, not just once completed. A signatory who has already signed did so against a specific set of terms — letting those terms move before every signatory has signed would silently invalidate signatures already collected. Every other field (signatures arriving, parties, etc.) stays mutable while `agreementStatus` is `SIGNING`, since that's the whole point of the status.
- Once `agreementStatus` is `COMPLETED` or `SUPERSEDED`, freezes the **entire record** — `data`, `signatures`, `agreementParties`, `attachments`, `historyEntries`, `references`, `metadata`, `template`, `uri`. Only `agreementStatus` itself (e.g. `COMPLETED -> SUPERSEDED`) and `state` stay writable. The guard (`assertAgreementRecordMutable` in `server/services/agreementService.ts`) is implemented as an allowlist of the fields that remain mutable rather than a list of frozen ones, so it fails safe for any field added to the model later. A resend of an unchanged value (e.g. same `data` with different key order) is a no-op, not a violation. Rejections throw a typed `AgreementRecordImmutableError` (409, code `AGREEMENT_RECORD_IMMUTABLE`, includes which field and status triggered it).
- Extends the same principle to **Templates**: once a template exists, its content is immutable (there's no first-class template versioning in this API, so a new version is a new row, not an edit in place — `assertTemplateContentImmutable`), and it can't be deleted while any agreement still resolves against it (`assertTemplateNotInUse`), since agreements — including frozen ones — render against a template's content.
- Both guards are wired through new optional `guardUpdate`/`guardDelete` hooks on the generic CRUD router (`server/handlers/crud.ts`), used only by the Agreement and Template routers — SharedModel is unaffected.
- `triggerAgreement` already only ever writes the `state` column, so trigger-driven runtime state updates keep working at every stage — verified, not changed.
- Mistakes on a signed/completed/superseded agreement (or an in-use template) are expected to be corrected by voiding/reinstantiating (or creating a new template), not by editing in place — this PR doesn't add those workflows, just the immutability guards.

## Test plan

- [x] `npx tsc -p . --noEmit` (server) — no type errors
- [x] `node --experimental-vm-modules node_modules/.bin/jest` (server) — 205/205 tests pass, including:
  - `services/agreementService.test.ts`: `assertAgreementRecordMutable` across DRAFT (fully open), SIGNING (`data` frozen, everything else open), COMPLETED/SUPERSEDED (everything frozen except `agreementStatus`/`state`), and no-op resends
  - `handlers/agreements.test.ts`: integration tests through the real `PUT /agreements/:id` route for each of the above
  - `services/templateService.test.ts` / `handlers/templates.immutability.test.ts`: template content-immutability and in-use delete guard, at both the service and route level
  - `handlers/crud.test.ts`: generic `guardUpdate`/`guardDelete` hook mechanism with synthetic guards, independent of Agreement/Template-specific logic
- [x] `npm run build` (root) regenerates `openapi.json`/`client/typescript/apap.ts`/`index.md`/`_site` cleanly from the updated model (the enum-typo fix was hand-applied instead of committing a full regeneration, since a fresh `npm run build` pulls unrelated upstream Concerto model drift into these generated files — kept the diff scoped to the actual fix)


niallroche:
Really careful change overall @mttrbrts I have a few observations — the staged freeze (data at `SIGNING`, whole record at `COMPLETED`/`SUPERSEDED`), the deny-by-default allowlist, the typed 409 errors, the no-op-resend courtesy, and the data-preserving `ALTER TYPE … RENAME VALUE` migration are all right, and `model/protocol.cto` + `db/schema.ts` agree after the rename. One correctness hole blocks it, plus two hardening items and two coordination flags.

**1. (Block) The freeze is reversible in two requests.** `agreementStatus` is itself in the mutable allowlist with no transition validation:

```ts
const MUTABLE_FIELDS_ONCE_FULLY_FROZEN = new Set(['agreementStatus', 'state']);
// ... if (MUTABLE_FIELDS_ONCE_FULLY_FROZEN.has(key)) continue;
```

So a client can `PUT` a `COMPLETED`/`SUPERSEDED` agreement back to `agreementStatus: 'DRAFT'` — allowed, since the only changed field is allowlisted — and on the next `PUT` the record is `DRAFT` and fully mutable again. Same path unlocks `data` via `SIGNING→DRAFT`. Nothing validates transitions (`concertoValidation` checks shape, not the state machine), so immutability falls in two calls. Needs an allowed-transition check (only forward: `DRAFT→SIGNING→COMPLETED→SUPERSEDED`, never a downgrade), plus a test for the revert case — currently the key missing one.

**2. (Significant) The guard is a non-atomic check-then-write.** The PUT path reads the row (`select … limit(1)`) and writes (`update … set(req.body)`) in separate statements with no transaction or `FOR UPDATE`, so concurrent PUTs (or a status flip racing an edit) can let a frozen-field write land after the guard saw a mutable state. `guardDelete` has the same TOCTOU. For a security-critical invariant the guard + write should share a transaction with a row lock.

**3. (Significant) Enforcement lives only in the HTTP router hook, not the mutation chokepoint.** `assertAgreementRecordMutable` is wired only in `agreements.ts`/`templates.ts`; the service mutators (`updateTemplate`, `deleteTemplate`, a future `updateAgreement`/direct `db.update`) don't call it. No live bypass today, but the moment a new MCP tool or `PATCH` route hits a service mutator the freeze is silently skipped. Move the assertion into the service update path.

**Coordination (de-conflict before the stack lands):**
- **#238 (LCP) writes `atrHash` on frozen records** — the deny-by-default allowlist rejects it unless `atrHash` is explicitly added to the mutable set. Please align with #238 on which.
- **#182 also fixes the `SIGNNG` typo.** `0002_fix_signing_enum_typo.sql` assumes the label is still `SIGNNG`; if #182 renames first this migration fails ("SIGNNG does not exist"), and `RENAME VALUE` can't be `IF EXISTS`. Only one typo migration can exist — de-dup at merge.

Minor: `state` is also plain-PUT-mutable on frozen records (broader than the "trigger-driven" justification), and template freeze rejects benign metadata edits (`description`/`keywords`) too. Both defensible — worth a conscious call.

mttrbrts:
Thanks for the thorough review — pushed `718d4d3` addressing all three substantive points.

**1. (Block) Reversible freeze — fixed.** `assertAgreementRecordMutable` now ranks the lifecycle (`DRAFT < SIGNING < COMPLETED < SUPERSEDED`) and rejects any `agreementStatus` change to a lower rank as a new `AgreementStatusTransitionError` (409, `AGREEMENT_STATUS_TRANSITION_INVALID`), checked first and unconditionally — before either freeze branch runs. Forward jumps (e.g. `DRAFT` straight to `COMPLETED`) stay allowed, same-status resends are a no-op. Added the missing test for every downgrade pair, the "revert bundled with a data change" case, all forward transitions, and an HTTP-level test proving the exact two-request exploit (`PUT agreementStatus: 'DRAFT'` on a `COMPLETED` row) is now rejected at the first request.

**2. (Significant) TOCTOU — fixed, scoped to guarded resources.** Guarded PUT/DELETE (Agreement, Template) now run the existence-check and the write inside one `db.transaction()` with `SELECT ... FOR UPDATE` locking the target row for the duration. `SharedModel` (no guard) keeps the simpler non-transactional path — nothing there for a race to bypass. Confirmed `db.transaction`/row locking is available on the existing `postgres-js` Drizzle handle (`server/index.ts`), so this didn't need any DB-layer changes, just crud.ts's PUT/DELETE handlers and the test doubles that exercise them.

**3. (Significant) Enforcement only at the HTTP hook — fixed for Template.** Moved `assertTemplateContentImmutable`/`assertTemplateNotInUse` into `updateTemplate`/`deleteTemplate` themselves (fetch existing row → guard → write), so the freeze holds at the mutation chokepoint regardless of caller. No REST route calls these functions today (the router still talks to the table directly, so this is additive — zero behavior change to current traffic), but it did surface that `templateService.test.ts`'s existing unit tests for these two functions were asserting a mutation the whole feature says shouldn't succeed; rewrote them to assert the (now correct) rejection, plus a no-op-resend success case and a `TemplateInUseError` case for delete. There's no equivalent standalone `updateAgreement`/`deleteAgreement` service function yet to wire the same way — noting that rather than inventing one speculatively.

**Also from Copilot's pass on this PR**: fixed the README's `PUT /templates/:id` example (was sending a partial body that 400s before ever reaching the guard) and added a comment on why `guardUpdate` intentionally runs after `transformRequest` (the guard needs to see the exact payload about to be persisted — no current caller combines the two, so it was inert either way).

**Left as conscious calls, not folded into this fix** (per your "worth a conscious call" framing):
- `state` stays writable via plain `PUT`, not only through `/trigger` — broader than the "trigger-driven" justification in the docstring. Restricting it is a separate design decision (would need to decide whether direct state correction via PUT is ever legitimate) rather than something to fold in here.
- Template's freeze still has no allowlist (`description`/`keywords` 409 too) — intentional per the existing "no versioning, a new version is a new row" design in this PR, not changed.

**Coordination flags (#238, #182)**: can't resolve these from this branch alone without visibility into where they've landed — flagging for whoever merges first to de-conflict (`atrHash` needs adding to `MUTABLE_FIELDS_ONCE_FULLY_FROZEN` if #238 lands after, and only one `SIGNNG`→`SIGNING` migration can survive if #182 lands first).

216 → 231 tests, `tsc --noEmit` clean.


---
_Generated by [Claude Code](https://claude.ai/code)_

niallroche (approved):
Re-reviewed `718d4d3` — all three are properly resolved, and I checked the logic in code, not just the summary:

- **Reversible status → fixed correctly.** `STATUS_RANK` (DRAFT<SIGNING<COMPLETED<SUPERSEDED) rejects any strictly-lower `toRank`, allows equal rank (no-op resend), allows forward jumps, and fails closed on an unknown status. Crucially it runs first and unconditionally, ahead of the frozen-field check — so a `COMPLETED→DRAFT` revert is now a 409, and the freeze can't be unlocked in two requests.
- **TOCTOU → fixed.** PUT and DELETE run the read+guard+write in one `db.transaction()` with `SELECT … FOR UPDATE`, and the delete guard runs its referential check inside the same lock via the `tx` handle.
- **Enforcement → fixed.** Guards moved into the service mutators, so the freeze holds at the mutation chokepoint regardless of caller.

Approving. Two coordination items remain, but they're downstream-rebase work, not blockers here (and match your note):
- **#238** must add `atrHash` to the mutable allowlist when it rebases on this (its post-completion write would otherwise be frozen out).
- Only one SIGNNG migration can survive — since this is ready and #182 is still a draft, #182 drops its typo fix and rebases on this.

One thing to confirm before merge: that the 15 new tests include the downgrade-rejection case (`COMPLETED→DRAFT` → 409) and the same-status no-op — the fix is correct, just want the regression pinned.

niallroche:
approved but there are conflicts to resolve @mttrbrts

mttrbrts:
Merge conflicts resolved (`73b52c3`) against main's #225/#226/#217 slice-3 REST-list-unification work that landed since your review — `listService`/`listAgreementsPaged`/`listTemplatesPaged` and the immutability guards touch adjacent-but-independent parts of the same files (`crud.ts`'s `CrudRouterOptions`, `agreements.ts`/`templates.ts`'s `buildCrudRouter` calls, both services' test files), so every conflict was additive — kept both sides, no logic changes. `npm install` picked up main's dependency-pin tightening too. `tsc --noEmit` clean, 261/261 tests pass (up from 205 purely from main's own new tests merging in, none of mine changed).

On confirming the specific regression tests — yes, both are pinned, at both layers:
- `COMPLETED → DRAFT` → 409: `services/agreementService.test.ts`, in the `agreementStatus cannot move backward` describe block, `it.each([['COMPLETED','DRAFT'], ['COMPLETED','SIGNING'], ['SUPERSEDED','COMPLETED'], ['SUPERSEDED','DRAFT'], ['SIGNING','DRAFT']])`. Plus an HTTP-level one in `handlers/agreements.test.ts`: `'rejects reverting agreementStatus COMPLETED -> DRAFT (would reopen the freeze)'`, and a second HTTP test, `'the COMPLETED -> DRAFT -> edit exploit is blocked at the first request'`, asserting the exploit path specifically.
- Same-status no-op: `'a same-status resend is not treated as a transition at all'` right after the downgrade cases in the same describe block (`agreementStatus: 'COMPLETED' → 'COMPLETED'`, doesn't throw).

Thanks for re-checking the logic rather than just the summary — appreciate the thoroughness on both passes. Agreed on the #238/#182 coordination being downstream, not blockers here.


---
_Generated by [Claude Code](https://claude.ai/code)_

PR #243, opened 2026-08-11
feat(mcp): paged resource URIs for apap://templates and apap://agreements (closes #217)

### LINKED ISSUE #217: [follow-up] Expose paged reads via MCP resource URIs for apap://templates and apap://agreements

JayDS22:
## Summary

Adds paged resource URIs for the two large-collection MCP resources on APAP:

- `apap://templates{?limit,offset}`: paged read of templates
- `apap://agreements{?limit,offset}`: paged read of agreements

Bare `apap://templates` and `apap://agreements` keep working and return page 1 (limit=100, offset=0), so existing readers are unaffected. The paged forms are progressive enhancement for clients that construct query-param resource URIs.

Closes #217.

## Design authority

Wire shape ratified by @niallroche in [#217 comment](https://github.com/accordproject/apap/issues/217#issuecomment-5237352992) after the URI-vs-tool proposal thread. This is the "closing PR of slice 3" Niall named there, sequenced after #225.

## Niall's guardrails

| Guardrail | Where addressed |
|-----------|-----------------|
| Pin SDK past UriTemplate ReDoS fixes (upstream `modelcontextprotocol/typescript-sdk` #965, #1334) | `@modelcontextprotocol/*@^2.0.0` in `server/package.json`. Server v2.0.0 ships the patched `UriTemplate` (`server/dist/src-CX2iR2pK.mjs:7220`): `match(uri)` runs `validateLength(uri, MAX_TEMPLATE_LENGTH, "URI")` before any regex, with `MAX_TEMPLATE_LENGTH=1e6`, `MAX_VARIABLE_LENGTH=1e6`, `MAX_TEMPLATE_EXPRESSIONS=1e4`. Non-exploded `?` operator resolves to bounded pattern `escapeRegExp(name) + "=([^&]+)"` (no `.+` backtracking). |
| Keep bare `apap://templates` and `apap://agreements` valid (page-1 fallback for clients that don't build query-param URIs) | Static resource registrations at `handlers/mcp.ts:298, 306`. Comment at `handlers/mcp.ts:314-315` documents the routing distinction between static (bare) and templated (query) registrations. |
| Use non-exploded query form (avoid `{var*}` ReDoS surface) | `'apap://templates{?limit,offset}'` at `handlers/mcp.ts:322` and `'apap://agreements{?limit,offset}'` at `handlers/mcp.ts:330`. Neither uses the exploded `*` modifier. |
| Server-side clamp on `limit` / `offset` | Already enforced by `listTemplatesPaged` / `listAgreementsPaged` in `services/templateService.ts` and `services/agreementService.ts` (landed in #225): `limit` clamps to 1-100, `offset` clamps to `≥0`. |

## Test plan

- [x] Unit: 42 tests in `server/handlers/mcp.test.ts` cover bare URIs, templated URIs, and edge cases (partial `?limit=` without offset, blank query values, values outside clamp range)
- [x] Passing locally on `@modelcontextprotocol/*@2.0.0` with cicero-core 2.x
- [ ] Manual smoke against `docker compose up` with `@modelcontextprotocol/inspector`
- [ ] Reviewer confirms the four guardrails above map to the cited lines

## Backwards compatibility

Bare-URI readers see zero change. Existing clients calling `resources/read` for `apap://templates` or `apap://agreements` continue to receive the first 100 rows without any client-side update.

## Sequencing

Depends on #225 landing first (slice 3 REST list unification) because `listTemplatesPaged` / `listAgreementsPaged` are the service functions this PR's paged resources call. Held per Niall's plan.

## Files changed

- `server/handlers/mcp.ts`: new `ResourceTemplate` registrations, kept static bare URIs
- `server/handlers/mcp.test.ts`: 42 tests covering all URI shapes and edge cases

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -153,6 +153,10 @@ jobs:
               END IF;
           END $$;
 
+          -- Return to the postgres superuser role before DROP: `rls_tester`
+          -- only has GRANT SELECT (not ownership) so a DROP as that role
+          -- fails with "must be owner of table _rls_smoke".
+          RESET ROLE;
           DROP TABLE _rls_smoke;
           SQL
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -110,10 +110,18 @@ jobs:
           psql -h localhost -U postgres -d postgres -v ON_ERROR_STOP=1 <<'SQL'
           CREATE TABLE _rls_smoke (id int PRIMARY KEY, owner text NOT NULL);
           ALTER TABLE _rls_smoke ENABLE ROW LEVEL SECURITY;
-          ALTER TABLE _rls_smoke FORCE ROW LEVEL SECURITY;
           CREATE POLICY p_owner ON _rls_smoke
               USING (owner = current_setting('app.user_id', true));
           INSERT INTO _rls_smoke VALUES (1, 'alice'), (2, 'bob');
+
+          -- Superusers bypass RLS regardless of FORCE ROW LEVEL SECURITY, and
+          -- the CI service runs as `postgres` (SUPERUSER). Demote to a fresh
+          -- NOSUPERUSER NOBYPASSRLS role for the assertions so the policy
+          -- actually applies. GRANT SELECT so the role can read the throwaway
+          -- table.
+          CREATE ROLE rls_tester NOSUPERUSER NOBYPASSRLS;
+          GRANT SELECT ON _rls_smoke TO rls_tester;
+          SET ROLE rls_tester;
 
           SELECT set_config('app.user_id', 'alice', false);
           DO $$
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -92,6 +92,62 @@ jobs:
           POSTGRES_URL: postgres://postgres:1baddeed@localhost:5432/postgres
         run: npx drizzle-kit push --config drizzle.config.ts
 
+      # Follow-up to Niall's #231 review: "postgres-18-smoke only applies the
+      # schema, not that RLS still isolates rows on 18." APAP itself does not
+      # yet enable ROW LEVEL SECURITY on any table (the auth chain in
+      # handlers/crud.ts is still commented pending the A2A + MCP auth PRs),
+      # so this step verifies that the *PG18 RLS feature itself* has not
+      # regressed by running a self-contained isolation test against a
+      # throwaway table. If PG18 ever ships a change that breaks RLS
+      # row-visibility semantics, this step catches it before APAP's own
+      # tables start using RLS.
+      - name: Verify PG18 RLS isolation semantics with a synthetic policy
+        env:
+          PGPASSWORD: 1baddeed
+        shell: bash
+        run: |
+          set -euo pipefail
+          psql -h localhost -U postgres -d postgres -v ON_ERROR_STOP=1 <<'SQL'
+          CREATE TABLE _rls_smoke (id int PRIMARY KEY, owner text NOT NULL);
+          ALTER TABLE _rls_smoke ENABLE ROW LEVEL SECURITY;
+          ALTER TABLE _rls_smoke FORCE ROW LEVEL SECURITY;
+          CREATE POLICY p_owner ON _rls_smoke
+              USING (owner = current_setting('app.user_id', true));
+          INSERT INTO _rls_smoke VALUES (1, 'alice'), (2, 'bob');
+
+          SELECT set_config('app.user_id', 'alice', false);
+          DO $$
+          DECLARE cnt int;
+          BEGIN
+              SELECT count(*) INTO cnt FROM _rls_smoke;
+              IF cnt <> 1 THEN
+                  RAISE EXCEPTION 'RLS regression: alice saw % rows, expected 1', cnt;
+              END IF;
+          END $$;
+
+          SELECT set_config('app.user_id', 'bob', false);
+          DO $$
+          DECLARE cnt int;
+          BEGIN
+              SELECT count(*) INTO cnt FROM _rls_smoke;
+              IF cnt <> 1 THEN
+                  RAISE EXCEPTION 'RLS regression: bob saw % rows, expected 1', cnt;
+              END IF;
+          END $$;
+
+          SELECT set_config('app.user_id', 'nobody', false);
+          DO $$
+          DECLARE cnt int;
+          BEGIN
+              SELECT count(*) INTO cnt FROM _rls_smoke;
+              IF cnt <> 0 THEN
+                  RAISE EXCEPTION 'RLS regression: unknown principal saw % rows, expected 0', cnt;
+              END IF;
+          END $$;
+
+          DROP TABLE _rls_smoke;
+          SQL
+
   publish:
     needs:
       - build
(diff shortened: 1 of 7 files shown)


JayDS22:
Self-review catch: the block comment above the paged `registerResource` calls referenced a `?? '100'` fallback that does not exist in the code. Fixed in `4f5aece`. Comment now accurately describes: `pageOpts` coerces to integers-or-undefined, service layer does the defaulting and clamping, and the SDK's UriTemplate regex today surfaces empty/absent params as ResourceNotFoundError before the callback runs (already pinned in the tests).

No code path change.

JayDS22:
Follow-up self-review pass caught a couple more before landing:

- `8a2423d`: `pageOpts` was accepting `parseInt`-friendly garbage silently. `parseInt('50abc', 10)` returns 50, `parseInt('50.5', 10)` returns 50, `parseInt('1e2', 10)` returns 100. All three now return `undefined` via a strict-integer regex + `Number.isSafeInteger` gate, so the service default applies instead of a partial parse. Trailing whitespace still tolerated for URL-encoded space cases. New test pins the behavior.
- **#244 filed** for the follow-up pagination completeness signal (`hasMore` / `_meta.nextUri` / `total` options). Not blocking this PR, but worth tracking as a client-ergonomics improvement while paging is fresh in review.
- Discovered while writing paged cache-hint tests: `ReadResourceResult` schema strips top-level `ttlMs` / `cacheScope` in transit to the client (pre-existing #199 / #201 shape issue, not this PR's fix scope). Paged callbacks inherit cache hints via the shared `getTemplates` / `getAgreements` path (`mcp.ts:325, 333` route to `mcp.ts:227, 258` where hints are spread), so the inheritance itself is by construction; the schema-transit issue is orthogonal.

CI green expected on both new commits.

niallroche:
Guardrails are all here — non-exploded `{?limit,offset}`, bare URIs preserved via the separate static registrations, and a strict `pageOpts` parser (`/^-?\d+$/` + `Number.isSafeInteger`, with `listTemplates` defaulting `opts.limit ?? 100` before clamping so `?limit=abc` can't NaN the query). Nicely done, and thanks for filing #244 for the completeness signals.

One should-fix before merge: `listTemplates` and `listAgreements` (the functions the paged resources call) have no `orderBy`:

```ts
return db.select().from(Template).limit(limit).offset(offset);
```

So `apap://templates?limit=50&offset=50` can repeat or skip rows relative to `offset=0` — unstable pagination. This is the exact issue #225 fixed for REST with `?? asc(Template.id)` in `listTemplatesPaged`; the MCP path routes through the un-fixed array siblings, so it reappears here. I checked #244 — it's scoped to `hasMore`/`nextUri`/`total` only, so the ordering fix has no other home and needs to land in this PR. Simplest fix is to mirror #225: add `.orderBy(asc(Template.id))` / `asc(Agreement.id)` to `listTemplates`/`listAgreements` so REST and MCP page over the same stable order.

Two smaller notes:
- The static-vs-template routing relies on the SDK matching the exact `apap://templates` registration before the `{?limit,offset}` template (which also matches an empty query). The 42 tests cover it, but a one-line comment naming that dependency would help the next reader.
- `console.log` in `getTemplates`/`getAgreements` — intentional structured logging, or debug residue? Might want the shared logger.

JayDS22:
Good catch on the ordering, pushed `3f1a36f` addressing all three:

1. **`listTemplates` / `listAgreements` stable orderBy**: mirrored #225's `asc(Template.id)` / `asc(Agreement.id)` default on the un-paged variants. REST (`listTemplatesPaged`) and MCP (`listTemplates` via the paged resource callback) now page over the same order. `mockDb` chain in `handlers/mcp.test.ts` updated to keep the fluent chain thenable after the new `.orderBy()` call. 94/94 tests pass across `templateService`, `agreementService`, and `mcp.test.ts`.

2. **Static-vs-template dispatch order comment**: added a forward-pointing note on both `apap://templates` and `apap://agreements` static registrations naming the SDK dispatch dependency (bare URI must be registered as a static resource so it wins the exact-string match before the `{?limit,offset}` template's regex is tried). The block comment on the paged registrations already explained the mechanism from the other direction; now both ends point at each other.

3. **`console.log` in `getTemplates` / `getAgreements`**: intentional structured-logging pattern, matches the rest of `handlers/mcp.ts` (`fetching_agreement`, `fetched_agreement_success`, `session_cleanup` etc. all use the same `console.log({ type: '...', ... })` shape). Not debug residue. Happy to switch the whole file to a shared logger in a follow-up if you want; keeping this PR scoped to #217 for now.

niallroche:
rebase on #245 and I can merge

JayDS22:
Rebased on **#245** (tip `846fe2b`), so the ordering is: **#245** commits (3) + this PR's commits (5) = 8 commits total on the branch. Diff on this PR page shows both stacks until **#245** merges to main; after that a `git fetch origin main && git rebase origin/main` is a no-op and the diff shrinks back to only this PR's changes.

Sequencing: merge **#245** first, then this one is ready.

Tests still pass locally: 96/96 across `handlers/mcp.test.ts`, `services/templateService.test.ts`, `services/agreementService.test.ts`.

niallroche (approved):
approving. The three notes from the last pass are all addressed:

- **Deterministic ordering** is now on both `listTemplates` *and* `listAgreements` (`.orderBy(asc(X.id))`), so MCP paging is stable and matches what #225 did for REST — no more repeat/skip across pages.
- **`pageOpts`** is strict (`/^-?\d+$/` + `Number.isSafeInteger`), and `listTemplates`/`listAgreements` default `?? 100`/`?? 0` before clamping, so junk like `?limit=abc` falls back cleanly with no NaN.
- Guardrails intact: non-exploded `{?limit,offset}`, bare `apap://templates`/`apap://agreements` preserved via the static registrations (with the dispatch-order comment), and the regression guard on ordering is a nice touch.

Two coordination notes, neither blocking:
- **Merge order:** this and #246 both edit `mcp.ts` (additive, mechanical conflict) — let's land #243 first and have #246 rebase.
- **#244** (page-completeness signals — `hasMore`/`nextUri`) stays as the tracked follow-up; correctly out of scope here.

Nice work @JayDS22

PR #245, opened 2026-08-11
chore(cleanup): restore #231 dep pins + PG18 RLS smoke + exact MCP SDK

JayDS22:
## What

Follow-up to #231 covering two items Niall flagged in his approving review that landed short.

## Why (honest disclosure of the miss)

When I rebased #231 on the current main yesterday to clear the `package.json` / `package-lock.json` conflicts before merge, I ran `git checkout --ours package.json package-lock.json` to take main's version of the dep files, then continued the rebase. In a rebase context `--ours` means the *target* (main), not the branch being rebased, so this dropped #231's actual dep-pin changes and kept main's older file. The other #231 changes (CRUD escape, test rewrites, upgrade script guards, the `postgres-18-smoke` job) landed correctly, but the dep pins did not. This PR restores them.

### 1. Restore the dep pins from #231

Confirmed in `server/package.json`:

- `"undefined": "^0.1.0"` removed (junk pkg from the SDK migration spike)
- `@accordproject/template-engine`: `^4.0.0` -> `4.0.0` (exact)
- `zod`: `^4.4.3` -> `4.4.3` (exact)

### 2. Pin `@modelcontextprotocol/*` exact `2.0.0` (Niall #231 non-blocking ask)

Was caret-ranged. Pinning exact ensures the floor includes the upstream UriTemplate ReDoS fixes ([modelcontextprotocol/typescript-sdk#965](https://github.com/modelcontextprotocol/typescript-sdk/issues/965), [#1334](https://github.com/modelcontextprotocol/typescript-sdk/issues/1334)) that matter for #243's resource read path where `UriTemplate.match()` runs on client-controlled URIs.

### 3. PG18 RLS-isolation smoke (Niall #231 non-blocking ask)

The `postgres-18-smoke` job today only verifies the schema applies via `drizzle-kit push`. APAP itself does not yet enable `ROW LEVEL SECURITY` on any table (the auth chain in `handlers/crud.ts` is commented pending the A2A + MCP auth PRs), so this new step verifies the *PG18 RLS feature itself* has not regressed by running a self-contained isolation test against a throwaway table:

```sql
CREATE TABLE _rls_smoke (id int PRIMARY KEY, owner text NOT NULL);
ALTER TABLE _rls_smoke ENABLE ROW LEVEL SECURITY;
ALTER TABLE _rls_smoke FORCE ROW LEVEL SECURITY;
CREATE POLICY p_owner ON _rls_smoke
    USING (owner = current_setting('app.user_id', true));
```

Then three assertions via `SELECT set_config('app.user_id', <name>, false)` + `count(*)`:
- `alice` sees 1 row
- `bob` sees 1 row
- unknown principal sees 0 rows

`FORCE ROW LEVEL SECURITY` prevents the `postgres` superuser from bypassing the policy. Any PG18 change that breaks RLS row-visibility semantics fails the job before APAP's own tables adopt RLS.

## Validation

- `npm install` clean; `undefined` package no longer in lock
- `npm test` 10 suites / 183 tests, all pass
- `@modelcontextprotocol/server` locked to `2.0.0` (exact) in `package-lock.json`

## Test plan

- [x] Local: `npm install` succeeds
- [x] Local: `npm test` all suites pass
- [ ] CI: matrix build stays green across Node 22.x + 24.x on all three OS
- [ ] CI: postgres-18-smoke passes the schema push AND the new RLS isolation step
- [ ] Reviewer confirms the dep pins now match what #231 was supposed to land

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -153,6 +153,10 @@ jobs:
               END IF;
           END $$;
 
+          -- Return to the postgres superuser role before DROP: `rls_tester`
+          -- only has GRANT SELECT (not ownership) so a DROP as that role
+          -- fails with "must be owner of table _rls_smoke".
+          RESET ROLE;
           DROP TABLE _rls_smoke;
           SQL
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -110,10 +110,18 @@ jobs:
           psql -h localhost -U postgres -d postgres -v ON_ERROR_STOP=1 <<'SQL'
           CREATE TABLE _rls_smoke (id int PRIMARY KEY, owner text NOT NULL);
           ALTER TABLE _rls_smoke ENABLE ROW LEVEL SECURITY;
-          ALTER TABLE _rls_smoke FORCE ROW LEVEL SECURITY;
           CREATE POLICY p_owner ON _rls_smoke
               USING (owner = current_setting('app.user_id', true));
           INSERT INTO _rls_smoke VALUES (1, 'alice'), (2, 'bob');
+
+          -- Superusers bypass RLS regardless of FORCE ROW LEVEL SECURITY, and
+          -- the CI service runs as `postgres` (SUPERUSER). Demote to a fresh
+          -- NOSUPERUSER NOBYPASSRLS role for the assertions so the policy
+          -- actually applies. GRANT SELECT so the role can read the throwaway
+          -- table.
+          CREATE ROLE rls_tester NOSUPERUSER NOBYPASSRLS;
+          GRANT SELECT ON _rls_smoke TO rls_tester;
+          SET ROLE rls_tester;
 
           SELECT set_config('app.user_id', 'alice', false);
           DO $$
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -92,6 +92,62 @@ jobs:
           POSTGRES_URL: postgres://postgres:1baddeed@localhost:5432/postgres
         run: npx drizzle-kit push --config drizzle.config.ts
 
+      # Follow-up to Niall's #231 review: "postgres-18-smoke only applies the
+      # schema, not that RLS still isolates rows on 18." APAP itself does not
+      # yet enable ROW LEVEL SECURITY on any table (the auth chain in
+      # handlers/crud.ts is still commented pending the A2A + MCP auth PRs),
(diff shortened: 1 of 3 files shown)


niallroche:
Quick follow-up on the SDK pin here @JayDS22 before this merges — pinning `@modelcontextprotocol/*` to exact `2.0.0` may lock us to a ReDoS-vulnerable version. The UriTemplate ReDoS (#965) was fixed in typescript-sdk PR #1363, **merged Jan 7, 2026** and shipped as a *patch* bump to `core`/`server` (backported to 1.x as #1365). So `2.0.0` is only safe if it was published on or after that date.

Can you confirm the resolved version with:

```bash
npm view @modelcontextprotocol/server versions --json
npm view @modelcontextprotocol/server@2.0.0 time   # is this ≥ 2026-01-07?

**Comment by JayDS22** [comment]
Good catch to check. Confirmed 2.0.0 is safe:

```
$ npm view @modelcontextprotocol/server@2.0.0 time --json | jq '."2.0.0"'
"2026-07-27T23:55:22.239Z"
```

All 5 sibling packages (`core`, `server`, `client`, `express`, `node`) at 2.0.0 published within the same second on 2026-07-27, which is ~6 months after the Jan 7 2026 ReDoS fix. The pre-release train also shows 10 alpha/beta releases from April through July 21, all post-fix, so the 2.0 line has carried the patched UriTemplate throughout.

Cross-check on the installed source at `node_modules/@modelcontextprotocol/server/dist/src-CX2iR2pK.mjs:7220` confirms the patched implementation:

```
const MAX_TEMPLATE_LENGTH = 1e6;
const MAX_VARIABLE_LENGTH = 1e6;
const MAX_TEMPLATE_EXPRESSIONS = 1e4;
```

Plus `match(uri)` calls `validateLength(uri, MAX_TEMPLATE_LENGTH, "URI")` before any regex work, and the `?` operator's non-exploded query form resolves to the bounded `escapeRegExp(name) + "=([^&]+)"` pattern (no `.+` backtracking).

On the exact-pin tradeoff (locking out future patch uptake): this matches the exact-pin convention #231 established for the other deps in this batch (`template-engine`, `zod`, `cicero-core`, `concerto-core`), on the "explicit pin so review can bisect against a known version" rationale. Happy to switch back to `^2.0.0` if the security patch uptake is worth breaking that convention for this specific dep, but the current 2.0.0 is a safe floor.

PR #246, opened 2026-08-12
Add MCP tool for agreement creation

stevenobiajulu:
## Summary

- add a generic `create-agreement` MCP tool
- create agreements through the shared `createAgreement` service that REST also calls
- validate the template relationship and Concerto agreement data before persistence
- return structured APAP service errors for invalid input and service failures
- keep APAP independent of any particular document renderer or vendor

## Why

This fills the agreement-creation gap in APAP's **MCP** surface. `POST /agreements` has always existed; what an agent could not do was create. It could list templates, read an agreement, convert it, and run its logic, but creation was the missing verb.

Clients can now create a standard APAP agreement and pass its Concerto data to output adapters such as the OpenAgreements DOCX bridge. The implementation is intentionally generic: OpenAgreements is a consumer of the APAP protocol, not a dependency of APAP.

## Why it is larger than one tool

The service extraction here follows the architecture proposed in #143 by moving agreement creation into a shared function called directly by REST and MCP. It extends the service-layer work already delivered in earlier PRs; it does not implement the whole RFC.

It came along because `create-agreement` could not be written honestly on top of the loopback. The first version accepted a bare template URL, which the Concerto validator behind `POST /agreements` rejects outright — the test was green only because it replaced `global.fetch` with a mock and never reached the validator. Reaching the real validator meant leaving the loopback. Creation now lives in `createAgreement(db, input, options?)`, which both `POST /agreements` and the tool call.

The executable `create-agreement` path no longer contains an internal HTTP loop and no longer reads `API_BASE_URL` or `APAP_API_AUTH_HEADER`; it calls `createAgreement` with the injected database directly, which closes the path where an MCP call borrowed the server's own `Authorization` header. Historical references to those identifiers remain in comments, docs and `server/.env_example` and could be cleaned up separately.

If you would rather review the service extraction on its own, I am happy to split it.

## Review follow-ups

**MCP schema strictness — fixed.** Copilot flagged that `inputSchema: AgreementCreateSchema.shape` dropped `.strict()`, and that was right. SDK 2.0's primary `registerTool()` overload takes a complete Standard Schema; the raw-shape form is deprecated and wraps the shape in a non-strict `z.object()`, so unknown fields were stripped before the callback ran and the service's strict `safeParse` never saw them. The tool now registers the complete `AgreementCreateSchema`. The advertised tool schema carries `additionalProperties: false`, and an unknown key is rejected before the service executes; both are covered by tests.

**Template URL parsing — tightened.** `normalizeTemplate` now parses candidate `http(s)` values with `new URL()` rather than matching a regex, so a malformed URL returns a typed `400` up front instead of failing later in Concerto validation, and the canonical form is what gets persisted. Non-`http(s)` values are still passed through unchanged.

**REST 422 coverage and changelog — added,** per @niallroche's review: a supertest case pinning `422 VALIDATION_ERROR` on `POST /agreements`, and `server/CHANGELOG.md` recording both contract changes.

## One open issue: `ap://` and #241

A bare `ap://name@version#hash` template URI is rejected today. Concerto accepts a relationship value only as the `resource:` scheme or with no scheme at all, and there is no branch for `ap://`, so it reaches the validator unchanged and fails with `Invalid URI scheme`. #241 puts exactly that value in an agreement's `template` field, so as written the two PRs do not compose.

A two-line fix that promotes `ap://` alongside `http(s)` passes the suite. I have not included it because the general version — promote any non-`resource:` scheme — makes `ftp://example.com/late-delivery.cta` succeed, and that rejection is deliberately asserted in the existing tests. Which URI schemes the create path should own looks like a decision worth making explicitly rather than inferring, so I would rather have the rule than guess it.

## Checks

- `npm run build` / `tsc -p . --noEmit` — exit 0
- 10 suites / 239 tests passing on the current head, rebased onto `main` after #243
- every new assertion confirmed able to fail against a deliberately broken handler

## Companion work

- OpenAgreements APAP template export and DOCX rendering: open-agreements/open-agreements#660
- Hosted signed-download bridge: UseJunior/openagreements-org-deploy#17


stevenobiajulu:
Pushed `d182132`, which reworks this PR after a review found the tool could not create the agreement its own description advertised.

**The core problem.** `create-agreement` accepted a bare template URL, but the Concerto validator behind `POST /agreements` rejects one — `Invalid URI scheme: https://…`. Only the canonical `resource:org.accordproject.protocol@1.0.0.Template#<url>` relationship form passes. The test was green because it replaced `global.fetch` with a mock and never reached the validator.

Both forms are now accepted: a plain absolute http(s) URL is normalised to the canonical form before validation, a canonical value passes through untouched, and a scheme-less identifier still reaches Concerto unchanged as it did before.

**Layering.** The tool POSTed back into this server's own REST API via `makeApiRequest`, which reversed the migration that `agreementService.ts` and `templateService.ts` document (`Replaces: makeApiRequest(...)`), and ignored the `db` handle injected into `getServer(db)`. Creation now lives in `createAgreement(db, input, options?)`, which both `POST /agreements` and the tool call. `makeApiRequest`, `API_BASE_URL` and `APAP_API_AUTH_HEADER` are removed — that also closes the path where an MCP call borrowed the server's `Authorization` header.

Also in scope: a structured zod input schema replacing `agreement: z.string()` (rejecting `id`/`templateHash`), typed `{ code, message, details }` failures via `serviceErrorToCallToolResult`, and two bugs the extraction surfaced — `concertoValidation` JSON-stringifies `data`/`state` in place and the inlined route persisted that mutated body, so rows carried `data` as a string while every reader treats it as an object; and a `23505` from the template cache was reported as a duplicate *agreement* URI, though `onConflictDoNothing` targets `Template.hash` and the separate `Template.uri` constraint names a different row.

**Two REST behaviour changes I'd like a view on**, both documented in `server/README.md`:

1. `data` must now be a JSON object rather than any JSON value. An agreement's data is a Concerto instance and needs a `$class`, but the previous table-derived schema did accept scalars.
2. Validation failures now use the repo's typed-error contract, so a Concerto rejection is `422 VALIDATION_ERROR` with a nested body rather than `400` with a flat one. This follows `globalErrorHandler`'s stated purpose, but it is a public endpoint.

**One thing deliberately left alone.** `HttpTemplateRetriever` has no allowlist, timeout, size cap or redirect policy, and attaches `EXTERNAL_TEMPLATE_TOKEN` to caller-chosen hosts. That is a pre-existing REST exposure which this tool gives MCP callers another route to. Hardening it is a security change to code this PR doesn't otherwise touch, so I'd rather have a maintainer ruling on whether MCP and REST share one trusted caller boundary than make that call in here.

Tests run the real Zod and Concerto validators against a real `.cta` fixture with only archive retrieval injected. Every new assertion was checked against a deliberately broken handler to confirm it can fail. Full suite: 203 passing.

stevenobiajulu:
Pushed `659a770` and marking this ready for review.

**One change since the last update.** `AgreementCreateSchema` restated the `agreementStatus` values as a literal list, which duplicated content that `db/schema.ts` generates from `model/protocol.cto`. It now reads `AgreementStatusType.enumValues`, so a status added or renamed in the model reaches the create path by regeneration instead of by someone remembering to edit a second list.

That matters for ordering against #240, which renames `SIGNNG` → `SIGNING` at source. With the derivation in place, #240 lands without touching this PR's create path — I verified by temporarily applying the rename to the generated schema and confirming `SIGNING` was accepted and `SIGNNG` rejected without any edit here.

One test change came with it: `agreements.test.ts` automocked `db/schema` wholesale, which left the derived enum with no values and turned every request body into a 400. The mock now keeps the pgEnum real and automocks the rest.

**Deliberately out of scope**, so reviewers don't have to guess whether they were missed:

- `HttpTemplateRetriever` hardening — pre-existing REST exposure, needs a ruling on whether MCP and REST share one trusted caller boundary.
- The crud router still mutates `req.body` through `concertoValidation` on PUT/PATCH — the same family as the bug fixed here, but that code is not part of this PR.
- `organization` has no column in the generated Agreement table. The route preserves the pre-refactor acceptance behaviour and nothing more; giving it a column is a schema decision.
- Tool-name casing (`create-agreement` vs `getTemplate`).

Separately, and not a change to this PR: while chasing the stale `.cta` fixture I found the RI cannot load **any** template in the public library. Our own fixture declares `cicero: ^0.25.0`, the newest hosted archives declare `^1.0.0`, and the runtime pins `cicero-core 2.1.1`. Relaxing the manifest guard on a `^1.0.0` archive loads it and constructs a processor fine, so it appears to be the version check rather than the format. That affects #241's cold-start path as much as this tool. I'll raise it at Wednesday's call rather than fold it in here.

Build clean, 203 tests passing, each new assertion confirmed able to fail against a deliberately broken handler.

stevenobiajulu:
Updated agreement creation to accept the OpenAPI-published JSON-string form for `data`, normalize it to an object before Concerto validation and persistence, and normalize object-encoding `state` strings without narrowing the existing state input surface.

Added service, REST, and MCP coverage for accepted inputs, rejection messages, persistence normalization, special-key parity, and MCP schema validation.

Verification:
- Build clean
- 10 suites / 217 tests passed
- Mutation-tested by removing normalization and confirming the relevant service, REST, and MCP tests fail

niallroche (commented):
This is in good shape — the service-layer factoring (real `createAgreement`, no HTTP loopback, one shared schema, typed 400/422/409 errors), the spec-shaped-JSON-string normalization, and the Drizzle-parameterized paths are all clean, and the `data`-must-be-object contract is now covered at the HTTP level (400 on `{not-json}` / `[1,2]`). Happy to approve with two small things first:

1. **Add a REST test that pins the 422.** The headline change here is that a Concerto validation failure returns **422** (vs the old 400), but there's no `request(app).post('/agreements')` asserting it — it's only exercised at the service layer. Since that status is now a public contract, please add one supertest asserting `422` + the typed error body on a semantically-invalid `data`. (The 400/data-object path is already well covered — this is just the other half.)
2. **Add a changelog / release-note entry for the contract change.** The README documents the 400→422 split and the object-only `data`, which is great, but a breaking REST change should also land in the changelog so downstream client authors see it.

Non-blocking: the `inputSchema: … .shape as any` cast (confirm the SDK conveys the object constraint to clients, else drop it) and the permissive `normalizeTemplate` regex (a `new URL()` would give a crisper up-front error) are fine as follow-ups.

Merge order note: this and #243 both touch `mcp.ts` (additive, mechanical conflict) — suggest #243 lands first and this rebases.

stevenobiajulu:
Thanks for the review — both blocking asks are now addressed, along with the two non-blocking points.

**REST 422 test** — added, asserting `422 VALIDATION_ERROR` and the typed body on `POST /agreements` for semantically-invalid `data`.

**Changelog** — `server/CHANGELOG.md` now records both contract changes: the `data` object rule and the 400/422 split.

**The `.shape as any` cast** — you were right to ask: it did not preserve strictness. With the old registration the advertised schema omitted `additionalProperties`, and the SDK stripped an unknown field before invoking the callback. `create-agreement` now registers the complete `AgreementCreateSchema`; the advertised schema includes `additionalProperties: false`, and unknown fields are rejected before the service executes. Both behaviours are covered by tests.

**`new URL()`** — taken. Candidate `http(s)` values are now parsed, malformed values return a typed 400 before Concerto validation, and the canonical form reaches the agreement insert. Other values still pass through normalization unchanged.

**Merge order** — #243 has landed and this head is rebased onto its merge commit. `pageOpts` and #243's paging tests are intact, and the `makeApiRequest` implementation is gone. Full suite passes: 10 suites, 239 tests.

One remaining question: a bare `ap://name@version#hash` template URI is currently rejected, so #246 and #241 do not compose as written. Promoting `ap://` into the canonical relationship form passes the full suite. I have not generalized that rule, because promoting every non-`resource:` scheme also admits `ftp://example.com/late-delivery.cta`, contrary to an existing rejection test. Should the create path explicitly recognize `ap://`, or should #241 supply the canonical relationship form?

PR #251, opened 2026-09-02
fix: normalize repeated filter query params

### LINKED ISSUE #249: Array-shaped filter values in defaultWhereClause bypass query normalization, cause 500
parseQueryParams uses getSingleQueryParam() to normalize repeated query params (Express parses ?page=2&page=3 as ['2','3']) — applied to page, limit, sortBy, and sortOrder, with test coverage in crud.test.ts (lines 547–560). ...filters isn't passed through the same normalization.

Repro: GET /templates?author=~a&author=~b → filters.author arrives as ['~a', '~b']. In defaultWhereClause, typeof value === 'string' is false for an array, so the whole string-handling branch (tilde/ILIKE, comparison operators) is skipped, and it falls through to sql\${column} = ${value}`. Postgres rejects text = text[]` (error 42883), and the error propagates as an unhandled 500.

Proposed fix: extend getSingleQueryParam (or equivalent normalization) across all keys in filters inside parseQueryParams, so values reaching defaultWhereClause are always normalized primitives — same pattern already used for page/limit/sortBy/sortOrder.

Happy to send a PR — small change (~5 lines) plus a test mirroring the existing repeated-param test for page/limit.

Src-Yash:
## Summary

Fixes #249.

Repeated filter query parameters could be parsed as arrays and passed
directly to the SQL query builder, resulting in invalid SQL and a 500
response.

This change normalizes filter values using the existing
`getSingleQueryParam()` helper so filters behave consistently with other
query parameters.

## Changes

- Normalize filter query parameter values in `parseQueryParams()`
- Use the first value when a filter is repeated
- Add a regression test for repeated filter parameters

## Testing

- `npm test -- handlers/crud.test.ts --runInBand` — 43/43 passed
- `npm test -- --runInBand` — 262/262 passed
- `npm run build` — passed
- `git diff --check` — passed

Note: `npm run lint` currently reports that no valid TSLint rules are
specified for TypeScript files.


niallroche (approved):
Looks good to me. Tiny optional test tweak: since #249’s repro was `?author=~a&author=~b`, it may be worth making this regression case use `['~Rob', '~Bob']` so the test documents that repeated filter params still preserve the case-insensitive filter operator path after normalization. Not a blocker.

Src-Yash:
Hi @niallroche, thanks for the review and approval. I’ve kept the current test unchanged since the suggested tweak was noted as optional/not a blocker. The required Unit Test workflows are currently awaiting maintainer approval. Could you please approve/run them when you get a chance? Thanks!

Issue #247, opened 2026-08-12
A2A wrapper design spec for review (spec-first)


JayDS22:
# A2A wrapper for accordproject/apap: design spec

Design proposal for adding an A2A (Agent-to-Agent) transport surface to the Accord Project Agreement Protocol (APAP) reference implementation alongside the existing MCP surface.

---

## Motivation

APAP already exposes agreements + templates over an MCP transport for AI-native consumers (Claude Desktop, Cursor, etc.). The next class of consumers is other agents: orchestrators, workflow engines, and autonomous agents in multi-agent systems. These do not speak MCP; they speak the A2A (Agent-to-Agent) wire protocol, a JSON-RPC 2.0 dialect for message-passing and task lifecycle management between agents.

Shipping A2A on APAP unlocks two concrete scenarios:

1. **Multi-agent orchestration.** A workflow engine coordinating multiple agents can delegate agreement operations to APAP as one of its agent participants without needing a bespoke bridge.
2. **Agentic payments.** The Accord Project's agentic payments direction requires a standard agent-to-agent surface where remote agents can trigger agreements, query task status, and receive settlement confirmations.

Adding A2A does not replace MCP; the two transports serve different client classes and share the same underlying services + data layer.

---

## TL;DR

- **What ships**: `POST /a2a` (A2A JSON-RPC endpoint) + `GET /.well-known/agent-card.json` (A2A discovery document) alongside the existing `POST /mcp`. Both transports call a shared `AuthAdapter` utility and reuse the existing services layer.
- **Architecture**: sidecar. A2A runs on its own route, not integrated into the MCP transport.
- **Deployment**: single Express process. `@a2a-js/sdk`'s `jsonRpcHandler` returns an Express `Router` that mounts on the existing app via `app.use('/a2a', ...)`. No second server.
- **Auth for v1**: symmetric HS256 shared-secret JWT via a bundled `JwtAdapter` reference. The `AuthAdapter` interface maps to the SDK's `UserBuilder` + `OwnerResolver` + `contextBuilder` hooks and stays extensible for asymmetric (RS256/JWKS), mTLS, or OIDC in future adapters.
- **Sequencing**: two PRs. **PR 1** ships A2A + shared `AuthAdapter` + `JwtAdapter` + `ApapAgentExecutor` + discovery card + tests. **PR 2** wires the same `AuthAdapter` into the existing `/mcp` and CRUD routes (auth currently commented out).

---

## Chosen architecture: sidecar

[...]

PR #253, opened 2026-09-14
feat(a2a): Add A2A protocol implementation

niallroche:
Implements PR 1 of the A2A design spec produced by @JayDS22 in accordproject/apap#247: an A2A Protocol v1.0 JSON-RPC endpoint running in the existing Express process alongside MCP and REST, calling the same shared services.

## What's included

- **Routes:** `POST /a2a` (JSON-RPC: `SendMessage`, `GetTask`, `ListTasks`) and a public `GET /.well-known/agent-card.json`, both built on `@a2a-js/sdk` 1.0.1.
- **Skills:** `list-templates`, `get-template`, `list-agreements`, `get-agreement` and `trigger-agreement`. `create-agreement` is left out until agreement creation lives in a shared service.
- **Authentication:** an open `registerAdapter(name, factory)` registry with two built-ins: `hs256` (demo-grade shared-secret JWT, which logs a warning in production) and `none` (development/test only). Adapters only see request headers and return a common `Principal` (`sub`, `orgId`, `roles`, `scopes`, `raw`). One adapter instance is created in `index.ts` so PR 2 can reuse it.
- **Authorization inside the services:** every shared template and agreement service takes a `PolicyContext { db, principal }` and calls `authorize(context, action, resource)` before touching the database. The reference policy requires the exact scope for each action (no wildcards from tokens). Policies can be replaced through `registerAuthorizationPolicy`. REST and MCP keep today's open behavior through a `DevelopmentPrincipal` until PR 2.
- **Configuration:** Zod-validated environment variables, checked at startup. `AUTH_ADAPTER` must be set in every environment, and production requires an HTTPS `PUBLIC_BASE_URL`.
- **Errors:** unknown skills, malformed input and insufficient scope end in `TASK_STATE_REJECTED`; failures during execution end in `TASK_STATE_FAILED`. Validation and authorization run *before* the task reports `TASK_STATE_WORKING`, so a refused request never looks like it started. Errors are serialized through an allowlist: APAP's own messages are kept, schema issues are reduced to the failing path and expected type, Concerto violations to the violated path and type, and wrapped template-runtime text is reported by code alone, so a caller never receives data its scopes do not cover. An error type can opt in to sending its own message with `ClientSafeError`. Unexpected internal faults return a generic `INTERNAL_ERROR`.
- **Task isolation:** the in-memory task store's owner key encodes the `(orgId, sub)` tuple as JSON, so principals whose fields could concatenate to the same string never share a task bucket.
- **Dev tooling:** `npm run a2a:dev` starts Postgres, applies the schema, builds, starts APAP, and smoke-tests the agent card and an A2A call.
- **Docs:** `docs/a2a.md`, `docs/auth.md` and `docs/a2a-release-notes.md`.

## Intentional differences from the spec

- **Missing or invalid credentials return HTTP 401** before JSON-RPC dispatch, instead of a failed task. Unauthenticated requests never reach the executor.
- **The agent card ignores `X-Forwarded-Host`.** Its URL comes only from `PUBLIC_BASE_URL`, which prevents card poisoning.
- **The built-in adapter is named `hs256`**, following the naming agreed in the #247 comments.

## Known and accepted limitations

- **The in-memory task store is unbounded.** It's the SDK's `InMemoryTaskStore` with no retention or size limit. Every task, including its request history and result artifact, stays in memory until the process restarts, so memory grows with request volume even on a single instance, and tasks aren't shared across replicas. This is knowingly accepted for this release: the spec defers a custom `TaskStore`. A persistent, bounded store is needed before long-running or horizontally scaled deployments.
- **`trigger-agreement` isn't idempotent.** A client retry can run the same trigger twice. This is documented in `docs/a2a.md`; callers should avoid blind retries after an ambiguous timeout.
- **No in-process rate limiting.** The spec leaves this to the reverse proxy.

## Blockers for PR 2

- The generic REST CRUD get/create/update/delete handlers in `handlers/crud.ts` still query `res.locals.db` directly. They must move behind the shared services and `authorize()` before REST authentication is turned on; otherwise those paths would skip the policy. This is marked in `crud.ts`, `docs/auth.md` and the release notes.

## Upgrade notes

This release can break existing deployments. See `docs/a2a-release-notes.md`. In short:

- `AUTH_ADAPTER` is now required in every environment.
- `none` is only accepted with `NODE_ENV=development` or `NODE_ENV=test`.
- `hs256` requires `A2A_JWT_SECRET` (at least 32 characters), `A2A_JWT_ISSUER` and `A2A_JWT_AUDIENCE`.
- Node.js 22.12 or newer is required.

## Testing

- `npm run build` passes.
- `npm test` passes: 18 suites, 303 tests. Use `npm test`; it supplies the `--experimental-vm-modules` flag that the trigger tests need on Node 22.
- `pretest` loads the real SDK runtime outside Jest, because Jest maps `jose` to v5.
- New coverage: JWT verification (expiry, audience, issuer, signature, `alg: none`, missing `exp`/`sub`, `nbf`), config fail-closed rules, the adapter registry, executor event order and error mapping, scope denial over HTTP, task isolation by principal and org, the 1 MB body limit, agent-card fallback, and the central policy.
- Regression tests for the three review findings: a colliding `(orgId, sub)` pair cannot read the other principal's tasks; refused requests never publish `WORKING`; and template-runtime text, submitted values and internal fault text never reach the caller. Each was confirmed to fail without its fix.
- `npm run lint` exits cleanly but checks nothing, because TSLint has no rules configured. That's the same on `main`.

## Reviewing

Reviewing commit by commit is easiest:

1. `e542307` A2A transport, auth adapters, executor, agent card
2. `c63a7be` one-command dev runner
3. `a486d7f` hardening and the `PolicyContext` refactor (touches every service, REST and MCP)
4. `6d418d4` docs, release notes, test-mock cleanup
5. `48ec8df` return service error details to A2A callers
6. `dba602a` close the error leak, the task-key collision, and premature `WORKING`


server/a2a/executor.ts
```diff
@@ -0,0 +1,296 @@
...
+            ...(Array.isArray(issue?.path) && { path: issue.path.join('.') }),
+            ...(typeof issue?.code === 'string' && { code: issue.code }),
+            ...(issue?.code === 'invalid_type' && { expected: issue.expected, received: issue.received }),
+            ...(issue?.code === 'unrecognized_keys' && { keys: issue.keys }),
+        })),
+    };
+}
+
+// Concerto violations arrive as a raw exception message that can quote the
+// submitted instance, so only the JSON path and expected type are forwarded.
```
mttrbrts, line 118: This should be an upstream fix in Concerto

server/README.md
```diff
@@ -98,6 +100,38 @@ npm start
...
+`GET /.well-known/agent-card.json`. See [docs/a2a.md](docs/a2a.md) for request
+examples, supported skills, and configuration, and [docs/auth.md](docs/auth.md)
+for the reference JWT adapter's security posture. Existing deployments should
+also read the [A2A release notes](docs/a2a-release-notes.md) before upgrading.
+
+To start PostgreSQL, initialize and build the server, run A2A smoke tests, and
+keep the development server running:
+
+```bash
+npm run a2a:dev
```
mttrbrts, line 129: Why do we want a separate launch script? Presumably this is just one process with both APAP, MCP + A2A endpoints?

mttrbrts (approved):

JayDS22:
Overall LGTM as spec owner. Sidecar shape, AuthAdapter interface, and PolicyContext refactor all match #247. Three hardening commits address leaks that were latent in the original spec threat model; the WORKING-order invariant, error-text allowlist, and task-key JSON encoding are cleaner than what I had written in #247.

Two intentional deviations I want to accept and roll back into the spec:

1. Agent card URL from PUBLIC_BASE_URL only. Safer than the spec's X-Forwarded-Host reflection, which was a card-poisoning primitive.
2. HTTP 401 pre-dispatch for missing auth. Correct per RFC 7235 and consistent with the "refused requests never look like they started" invariant.

One request before merge: consider gating the DevelopmentPrincipal policy bypass on NODE_ENV in addition to config.ts. Two locks on the "all authorizations bypass" path are cheap.

Follow-up comment will carry the smaller findings on middleware order, agent card fallback path, and adapter registry ergonomics.

Blocker note: DCO signoff missing on all six commits. `git rebase --signoff HEAD~6` should clear it.

server/a2a/executor.ts
mttrbrts: Tracked upstream in accordproject/concerto#1325

server/README.md
niallroche: Agreed @mttrbrts , parsing message text isn't where this belongs. I've added a comment pointing at [accordproject/concerto#1325](https://github.com/accordproject/concerto/issues/1325) so it gets deleted once structured validation errors land. Until then A2A forwards only the extracted path and expected type, never the raw message, since that can quote the submitted instance.

niallroche:
hanks Jay. Taking the three points in turn.

`DevelopmentPrincipal lock` — added, though deliberately not inside `ScopeAuthorizationPolicy. REST` and MCP still reach the services through createLegacyPolicyContext (a DevelopmentPrincipal) until PR 2, so gating the bypass on `NODE_ENV` there would make every production REST and MCP call fail with `INSUFFICIENT_SCOPE` — exactly the breaking change PR 1 is meant to avoid. The two locks are instead:

the `none` adapter factory refuses to construct when `isProduction`, and
the transport rejects any unauthenticated principal with HTTP 401 under `NODE_ENV=production`, including one returned by a custom adapter.

That gives two independent locks on the A2A bypass path with REST and MCP behaviour unchanged. Once PR 2 moves those surfaces onto real principals the policy-level gate becomes safe, and I'd add it then. Say so if you'd rather have it in the policy now and accept the REST/MCP consequence.

Deviations — agreed on both. They're written up in docs/a2a.md and the release notes, so folding them into #247 should be close to copy-paste.

DCO — small correction: two of the six commits were missing the sign-off (a486d7f, 6d418d4); the other four had it. A remediation commit turned out not to work here, since the app only honours those when a repo opts in via .github/dco.yml, so I rebased with `--signoff` and force-pushed. DCO is green and the approval survived. The shas all changed, so earlier inline comments now point at outdated commits.

Also cleared CodeQL's missing-rate-limiting alert: /a2a now has an in-process limiter ahead of authentication (120 requests/minute per address by default, tunable via `A2A_RATE_LIMIT_MAX / A2A_RATE_LIMIT_WINDOW_MS`), with the edge proxy still the first line of defence. Worth reflecting in the spec, which had deferred rate limiting entirely.

Two invariants worth carrying into #247 as well: refused requests never publish `WORKING`(there's a test asserting it), and the task-owner key is `JSON.stringify([orgId ?? null, sub])` rather than a string join.

niallroche:
I added an in-process limiter sits ahead of authentication on '/a2a' (120 requests/minute per client by default, tuned with 'A2A_RATE_LIMIT_MAX and A2A_RATE_LIMIT_WINDOW_MS'), answering '429 RATE_LIMITED'. It keys on the client address, so deployments behind an ingress must set 'TRUST_PROXY'; production warns when it is unset.

